PR #17, opened 2014-06-11
errorwriter errors

mikeyc7m:
errcontext need to be an array, otherwise the errorwriters throw an error.


dhensby:
@nyeholt this can be closed too :)

tractorcow:
Thanks for fixing this @dhensby

PR #32, opened 2015-03-19
BUG Two processes can potentially start the same job due to race cond

halkyon:
Check the affected rows when setting JobStatus via DB::query() instead
of write(). If nothing was affected, then bail out as there's nothing to be done.

### DIFF
--- a/code/services/QueuedJobService.php
+++ b/code/services/QueuedJobService.php
@@ -291,7 +291,7 @@ class QueuedJobService {
 	 * @param QueuedJobDescriptor $jobDescriptor
 	 *			The Job descriptor of a job to prepare for execution
 	 *
-	 * @return QueuedJob
+	 * @return QueuedJob|boolean
 	 */
 	protected function initialiseJob(QueuedJobDescriptor $jobDescriptor) {
 		// create the job class
@@ -302,7 +302,6 @@ class QueuedJobService {
 			throw new Exception("Implementation $impl no longer exists");
 		}
 
-		// start the init process
 		$jobDescriptor->JobStatus = QueuedJob::STATUS_INIT;
 		$jobDescriptor->write();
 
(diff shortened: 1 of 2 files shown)


code/services/QueuedJobService.php
```diff
@@ -388,8 +401,20 @@ public function runJob($jobId) {
...
 				$jobDescriptor->JobRestarted = date('Y-m-d H:i:s');
 			}
-			
-			$jobDescriptor->JobStatus = QueuedJob::STATUS_RUN;
+
+			// write the status and determine if any rows were affected, for protection against a
+			// potential race condition where two or more processes start the same job at once.
+			// This deliberately does not use write() as that would always update LastEdited
+			// and thus the row would always be affected.
+			DB::query(sprintf(
```
mateusz: I don't think we need to modify this code here. The only way for us to actually "grab" the job is to go through the `initialiseJob`, which already has the race condition prevention. There is another potential source of problems down below though: there is a `write` in the "catch" block - if the "second" job encounters an exception in the `initialiseJob`, it will rewrite the status as STATUS_BROKEN, even though the "first" job might be merrily running along.

I think I'd suggest to refactor this:
- separate the UPDATE (the "grab" code) out into a separate function, and call it "grabJob" or something like that.
- pull it out of the try block. This way if the process fails to "grab" the job, it will just exit, instead of trying to do writes.

halkyon: @mateusz Thanks for that. I agree, and have updated the PR.

nyeholt:
I'm starting to feel like the responsibility of what QueuedJobService is doing is starting to get beyond what it should be doing, and that some of this 'queue' management should instead be done by the QueueHandler, which for other implementations (Gearman, SQS) would be handled by those queue runners internally. I'm happy to accept this as-is as a fix for now - but will make a not to look at this when I dig at the gearman implementation a bit more. Will check for more comments in the morning otherwise happy for it to go through.

PR #38, opened 2015-05-07
API Improve queuedjobs admin

tractorcow:
- Admin can delete or edit jobs
- Non-admins can view jobs as readonly in the CMS


tractorcow:
@nyeholt do you want to have a look at this? do you think this improves the usability of the admin from a debugging point of view?

This mostly came out of "ugh, my jobs are breaking and I don't know why" day.

nyeholt:
Yeah, not opposed to it, you're definitely not the only one who's had days like that :)

tractorcow:
Ah thanks mate. :D It's definitely going to make a few ppl in ops happy.

PR #39, opened 2015-05-11
BUG Fix dropdown not selecting correct status

tractorcow:
Because DropdownField needs an associative array not an indexed one.


tractorcow:
Cheers @nyeholt :D

PR #42, opened 2015-05-13
BUG Fix memory usage safety net from crashing on some sites

### LINKED ISSUE #41: OOM errors leave queuedjobs unrecoverable
There is memory protection in queuedjobs, but it's more of a mechanism for reducing the chance of this happening rather than recovering from such a state.

What normally happens is that the job will have JobStatus = 'Running', and when the job detects that memory is low it sets the job to 'Waiting' and restarts.

If this catch fails to trigger then the job will be left with status = waiting and there won't be any indication that the queue has failed (unless you explicitly check the server logs for the error message).

tractorcow:
Fixes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,30 @@ _must_ detect whether they're present or not before using them. See [this issue]
 and [this wiki page](https://github.com/silverstripe-australia/silverstripe-queuedjobs/wiki/Defining-queued-jobs) for 
 more information
 
+## Memory limit configuration
+
+By default this task will run until it comes within 12MB of the memory limit configured by PHP, giving the application
+room to safely shutdown and restart.
+
+In some cases the buffer can be adjusted to be more robust (increased) or more performant (decreased). To adjust this
+buffer set the following config.
+
+
+	:::yaml
+	# Reduce memory buffer to 10mb
+	QueuedJobService:
+	  memory_buffer: 10485760
+
+
+This limit can be completely bypassed by setting a hard limit on the exact memory, thus ignoring php_ini('memory_limit').
+
+
+	:::yaml
+	# Force memory limit to 256mb
+	QueuedJobsService:
+	  memory_limit: 268435456
+
+
 ## Indexes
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -112,6 +112,30 @@ _must_ detect whether they're present or not before using them. See [this issue]
 and [this wiki page](https://github.com/silverstripe-australia/silverstripe-queuedjobs/wiki/Defining-queued-jobs) for 
 more information
 
+## Memory limit configuration
+
+By default this task will run until it comes without 12MB of the memory limit configured by PHP, giving the application
```
assertchris: This sentence is confusing. Do you mean it will stop before it has 12MB left, or will work until it has used 12MB. And do you mean 128MB?

README.md
```diff
@@ -112,6 +112,30 @@ _must_ detect whether they're present or not before using them. See [this issue]
...
+room to safely shutdown and restart.
+
+In some cases the buffer can be adjusted to be more robust (increased) or more performant (decreased). To adjust this
```
assertchris: Will decreasing buffer size really increase performance?

README.md
tractorcow: If php has a memory limit of 132MB, and we have left the buffer at 12mb, it will continue executing until it uses 120mb (132 - 12).

It's confusing because I should have said "within" not "without"... I'll fix.

README.md
tractorcow: It means it'll restart fewer times, although it'll come closer to the memory hard limit set by php.

assertchris: Fair enough.

README.md
assertchris: That makes sense. Just that change then, and I would be happy to merge.

tractorcow:
I fixed my typo. :D

tractorcow:
@nyeholt any chance you could have a look at this today Marcus? :P

tractorcow:
You know, this doesn't actually fixe #41 after testing. I'm just going to close this until I can find a working fix.

PR #43, opened 2015-05-20
API Better restarting of stalled jobs (when StepsProcessed == 0)

tractorcow:
BUG Better memory detection
BUG Correct usage of configurable options in QueuedJobService
BUG Corrected "ResumeCount" to "ResumeCounts"

Another potential solution to https://github.com/silverstripe-australia/silverstripe-queuedjobs/issues/41

@nyeholt would you be able to please review my logic? I've written up a complete unit test covering job restart and I believe that my logic is correct, but it deserves some critique.

I have replaced the requirement that StepsProcessed must be > 0 in order to be restartable, and instead changed LastProcessedCount to default to -1 (and is set to 0 during checkJobHealth on jobs that are suspected of requiring restart). This means that although jobs won't restart immediately, if no work is done on them they will eventually restart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,28 @@ _must_ detect whether they're present or not before using them. See [this issue]
 and [this wiki page](https://github.com/silverstripe-australia/silverstripe-queuedjobs/wiki/Defining-queued-jobs) for 
 more information
 
+Ensure that notifications are configured so that you can get updates or stalled or broken jobs. You can 
+set the notification email address in your config as below:
+
+
+	:::yaml
+	Email:
+	  queued_job_admin_email: support@mycompany.com
+
+## Memory limit configuration
+
+By default this task will run until either 128mb or the limit specified by php_ini('memory_limit') is reached.
+
+You can adjust this with the below config change
+
+
+	:::yaml
+	# Force memory limit to 256 megabytes
+	QueuedJobsService:
+	  # Accepts b, k, m, or b suffixes
+	  memory_limit: 256m
+
+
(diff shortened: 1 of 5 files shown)


tractorcow:
Ah, results of testing this new code were excellent. I was able to run the process on a low-memory instance without the jobs completely failing. I had a few alerts, but no jobs were left stuck.

I'm going to add yet one more tweak to this; An option to set a max execution timeout, after which the job will be forced to restart.

tractorcow:
I've added this optional feature as a separate commit.

tractorcow:
Well, testing of the new feature didn't reveal anything amazing. The module still worked about the same; I'll leave it up to you @nyeholt if you think it should go in or not. :P

tractorcow:
@sminnee I have no idea how, but yes it can get to that point. In my tests I've seen cases where yes it restarts due to memory consumption being high.

nyeholt:
All looks reasonable to me - and you've made a few refactorings that I've had on my "need to clean that up" list for a while which is nice :D. There'll still be cases when OOM could still hit during a job's process(), but until there's a nice catch for a Fatal error, not much that can be done about that. 

> @sminnee I have no idea how, but yes it can get to that point. In my tests I've seen cases where yes it restarts due to memory consumption being high.

I can't see Sam's comment so don't have any context around this, but I presume it's nothing breaking anything :)

tractorcow:
I think he added then deleted his comment; It came through in my email :D

Regardless, thanks for the merge!!

PR #47, opened 2015-07-28
Added additional core engine

tractorcow:
Squashed and rebased version of https://github.com/open-sausages/silverstripe-queuedjobs/commits/feature/add-additional-core-engine

### DIFF
--- a/_config/queuedjobs.yml
+++ b/_config/queuedjobs.yml
@@ -7,11 +7,21 @@ Injector:
   QueuedJobService:
     properties: 
       queueHandler: %$QueueHandler
+  DefaultRule:
+    class: 'AsyncPHP\Doorman\Rule\InMemoryRule'
+    properties:
+      Processes: 1
+      MinimumProcessorUsage: 0
+      MaximumProcessorUsage: 100
+  ProcessJobQueueTask:
+    properties:
+      DefaultRules:
+        - '%$DefaultRule'
 
 ---
 Name: gearman_queue_settings
 Only:
-    moduleexists: gearman
+  moduleexists: gearman
 ---
 Injector:
   GearmanQueueHandler:
--- a/code/dataobjects/QueuedJobRule.php
+++ b/code/dataobjects/QueuedJobRule.php
@@ -1,0 +1,159 @@
+<?php
+
+use AsyncPHP\Doorman\Rule;
+
+/**
+ * @property int $Processes
+ * @property string $Handler
+ * @property float $MinimumProcessorUsage
+ * @property float $MaximumProcessorUsage
+ * @property float $MinimumMemoryUsage
+ * @property float $MaximumMemoryUsage
+ * @property float $MinimumSiblingProcessorUsage
+ * @property float $MaximumSiblingProcessorUsage
+ * @property float $MinimumSiblingMemoryUsage
+ * @property float $MaximumSiblingMemoryUsage
+ */
+class QueuedJobRule extends DataObject implements Rule {
+	/**
+	 * @var array
+	 */
+	private static $db = array(
+		'Processes' => 'Int',
+		'Handler' => 'Varchar',
+		'MinimumProcessorUsage' => 'Decimal',
+		'MaximumProcessorUsage' => 'Decimal',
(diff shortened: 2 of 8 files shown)


nyeholt:
Am just copying my comment through from the previous PR that got merged against the sausage branch (haha never thought I'd write those words together in a sentence)

The scope of this change makes me think it's probably time to properly abstract this part out of the base QueuedJobService. At present there's the concept of a 'queueHandler', that is currently only used when adding new jobs. My gut feel is that this should/could also be used for the 'process' part (or introducing a new dependency), and having this logic encapsulated in a DI over-ridable manner. So it'd be something like

```php
class QueuedJobService {

	/**
	 * Called to add a job to the queue. and to process jobs on the queue
	 */ 
	public $queueHandler;

	/** 
	 * OR as a separate entity itself
	 */
	public $queueProcessor;


	/**
	 * Start a job (or however the queue handler determines it should be started)
	 * 
	 * @param JobDescriptor $jobDescriptor
	 * @param date $startAfter
	 */
	public function startJob($jobDescriptor, $startAfter = null) {
		if ($startAfter && strtotime($startAfter) > time()) {
			$this->queueHandler->scheduleJob($jobDescriptor, $startAfter);
		} else {
			// immediately start it on the queue, however that works
			$this->queueHandler->startJobOnQueue($jobDescriptor);
		}
	}

	/**
	 * Start a job (or however the queue handler determines it should be started)
	 * 
	 * @param JobDescriptor $jobDescriptor
	 * @param date $startAfter
	 */
	public function startJob($jobDescriptor, $startAfter = null) {
		if ($startAfter && strtotime($startAfter) > time()) {
			$this->queueHandler->scheduleJob($jobDescriptor, $startAfter);
		} else {
			// immediately start it on the queue, however that works
			$this->queueHandler->startJobOnQueue($jobDescriptor);
		}
	}
	

	public function processJobQueue() {
		$this->queueHandler->processQueue($name, $this?);
		$this->queueProcessor->processQueue($name, $this?);
	}
}

```

A couple of benefits being that 

- the queuedjobs module doesn't need to depend on doorman directly
- the messy aspects of processing the queues are abstracted more appropriately, encouraging replacement
- something else I can't recall but I've got to run to a meeting so can't type any more :D 


Anyway - the general functionality I'm okay with, but I think the implementation approach has some scope for improvement in a general sense. Though I accept that it's possible that no-one has the time to do so just at the moment :)

assertchris:
@nyeholt I think that's a great idea. Do you think it should be in the same feature branch or a new one?

nyeholt:
Depends on the urgency for getting this into the base module :) If you've got the time to do the generalised refactoring, then I don't see a problem with it being in the same feature branch, as there'd be a few changes to how the doorman stuff hooks in (likely being able to replace the if / else stuff in ProcessJobQueueTask with a  much simpler approach).

assertchris:
I think there are full text search improvements that could benefit from this work, so maybe the refactoring could be part of another story. Shortly after this but before tagging the next stable release...

@tractorcow what do you think?

tractorcow:
I think it's fine; @assertchris will just have to refactor out the doorman specific stuff from the method switch into a separate subclass that's DI into the queue processor.

I think if we do make this change it makes sense to make it now, rather than later.

The way that ProcessJobQueueTask::run should work would be better with this DI pattern than a set of hard-coded methods.

tractorcow:
Well, I had a go at doing this, and it looks like 90% of the QueuedJobService would need to be refactored out into a new dependency.

I want to un-say what I agreed with and say that this is going to be too much work for us to fit in. =( It would be nice but sadly we don't have the time.

Can we just merge as-is? :P

tractorcow:
Will close and re-open soon.

tractorcow:
@nyeholt can you look at https://github.com/open-sausages/silverstripe-queuedjobs/pull/8 and tell me if that resolves your issue? :D

nyeholt:
Yep cool :)

nyeholt:
Which also means - am happy for this to be merged back when you're ready to Damian

tractorcow:
Nice; Once my team has approved and merged the change I'll squash and re-open this one.

PR #49, opened 2015-07-28
remove tailing spaces in services

spekulatius:


nyeholt:
There was a recent set of changes that got merged in that have caused a bunch of conflicts with this - is there a script / automated process you run to do these cleanups? Would be good to see if I can hook them into my general process :)

spekulatius:
Thanks for your reply :)

Its just a shell script which isn't really advanced. Is there a preference for a tool to run this kind of checks? Symfony uses a tool which hooks into their pull requests and checks the codestyle...

For now I just rebased the branch.

tractorcow:
Oh no....

https://github.com/silverstripe-australia/silverstripe-queuedjobs/commit/2780e801723bf169d6bee63f549d0a752623a443?w=1

Broke queuedjobs =(

spekulatius:
oh sh*t. sorry! I should have been more careful. :(

nyeholt:
Odd, I remember seeing that change in the initial changeset, thus the comment about the other upstream changes happening, but not in the rebase. I wonder if the phone UI wasn't showing the full changeset and just the new commits when I re-reviewed.

tractorcow:
You need to use ?w=1 when viewing the diff to check for non-whitespace changes. Normally on these kinds of PR I use this to prove that there are no functional alterations. :)

spekulatius:
That is a good thing. I didn't know this. Thanks @tractorcow :)

PR #51, opened 2015-07-29
Added additional core engine

tractorcow:
Rebased from https://github.com/silverstripe-australia/silverstripe-queuedjobs/pull/47

### DIFF
--- a/docs/en/configuring-multi-process-execution.md
+++ b/docs/en/configuring-multi-process-execution.md
@@ -1,0 +1,56 @@
+# Configuring Multi-process Execution
+
+You can enable multi-process execution by selecting `doorman` as the engine:
+
+	:::yaml
+	---
+	Name: myqueuedjobsconfig
+	After: '#queuedjobsettings'
+	---
+	Injector:
+	  ProcessJobQueueTask:
+	    properties:
+	      TaskRunner: %$DoormanRunner
+
+
+By default, this will allow a single child process to complete queued jobs. You can increase the number of processes allowed by changing the default rule:
+
+
+:::yaml
+	---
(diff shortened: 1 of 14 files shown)


tractorcow:
@nyeholt I've squashed and included the refactor of the top level core engine. Please let me know if you like this one. :D

docs/en/configuring-multi-process-execution.md
```diff
@@ -0,0 +1,56 @@
...
+	Injector:
+	  ProcessJobQueueTask:
+	    properties:
+	      TaskRunner: %$DoormanRunner
+
+
+By default, this will allow a single child process to complete queued jobs. You can increase the number of processes allowed by changing the default rule:
+
+
+:::yaml
```
assertchris: `\t`

nyeholt:
Am just playing around with a couple refactorings that might be worthwhile - give me an hour or so  :)

tractorcow:
I've added another commit to this based on yours (with a few tweaks). Hope this suits. :D

nyeholt:
Ah yes, I missed that runJob bit. The circular dependency is a bit of an issue - I kind of lean towards pushing the service through as a parameter rather than a singleton, as the singleton call  is essentially the same as the wired circular dependency.

tractorcow:
Hi @nyeholt  just checking in.  Any further action you'd like from me on this PR? :)

nyeholt:
Oh! I Thought you'd merged it already :) I think I'll change the getService to pushing the service down as a parameter, but I don't have the time to do so right now so instead of holding things up I'll merge now and change later.

nyeholt:
Have tagged a new minor version (2.8.0) too.

PR #52, opened 2015-08-03
Order jobs by latest created

wilr:
When in the CMS as a user I usually care about more recent log messages than older ones. This is especially the case when reloading the page to watch for updates from the task. If it's on the last page, GridField currently doesn't remember the page.


tractorcow:
I was just thinking that today @wilr

Issue #46, opened 2015-06-18
Jobs run concurrently with the same signature


torleif:
I have several jobs that I don't want to run at the same time. Even though the job returns a static signature through the getSignature() function, they continue to run concurrently at the same time. 

I'm setting isComplete = true when it completes, and I'm stepping through the currentSteps and have set the totalSteps correctly.

nyeholt:
The job signature is only used when you create a new job, and only prevents the creation of a new job if there's already a 'new' job with that signature in place - ie if there's already a job running, it will let you queue that new job up even though they have the same signatures. 

Do you have multiple queue executors configured? Or is it that you only want the job to be added once the current one finishes? For the second case, I generally have the job create a new version of itself when it finishes.

torleif:
I have it creating another job after it finishes. The queue executors (I assume) are the default ones described in the readme.md (our sites are running on CWP)

if you run the job queue task multiple times via cli, even if there's a job being run, it will continue to spawn multiple instances of the same job. 

The behaviour should check if the same job is not currently being executed. It's very misleading to have one job 'running', even though there often are multiple jobs of the same one being ran.

nyeholt:
Hmm, something sounds amiss. The way the executors are set up is to 

* call ProcessJobQueueTask::process
* Look to see if the provided $queue parameter has any pending jobs via QueuedJobService::getNextPendingJob (doesn't run the job, just checks to see if there are any)
* If it does, call QueuedJobService::processJobQueue
* Retrieve the next job via processJobQueue 
* Execute that job
* Call getNextPendingJob again, and if found, execute - looping this way until all jobs from that queue are complete

getNextPendingJob has an explicit check that looks for jobs executing on $queue (ie the jobs have a status of either INIT or RUNNING) - if it finds an already running job on that queue it returns false (and the previously mentioned job loop does not execute. 

Now - one thing that does stand out as a possible issue is if, on your job completing and creating a new version of itself, it creates a new job and adds it, which will then get picked up by that processJobQueue loop - and create a new job, which immediately executes, and so on. Could that be what's happening? Or are you actually having two (or three / four / five) instances of the job executing concurrently?

torleif:
[...]

PR #61, opened 2015-09-11
DOC document job health check workaround

nyeholt:
Fix for #46 - documented a workaround for the jobqueue health check mistakenly marking a job
as being paused (ie set currentStep = -1).

Tweaked some code in QueuedJobDescriptor to ensure that only jobs with a
StepsProcessed count > 0 are considered as needing a 'restart' instead of a
'setup'

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,9 +144,15 @@ set the notification email address in your config as below:
 	Email:
 	  queued_job_admin_email: support@mycompany.com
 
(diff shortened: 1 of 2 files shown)


spekulatius:
Hello @nyeholt ,

I had a look at this. I would be happy to get this merged in.

Spek

PR #63, opened 2015-09-20
Adding phpdoc blocks and removing some unneeded lines

spekulatius:


code/extensions/ScheduledExecutionExtension.php
```diff
@@ -18,16 +20,21 @@ class ScheduledExecutionExtension extends DataExtension {
 		'ExecuteFree'			=> 'Varchar',
```
assertchris: Could you un-align these? The extra space isn't helping readability...

code/jobs/PublishItemsJob.php
```diff
@@ -45,7 +45,6 @@ public function getTitle() {
 	 *
```
assertchris: Could drop this line...

code/extensions/ScheduledExecutionExtension.php
spekulatius: Hello @assertchris ,

thanks for your quick response. I left them as they are because I personally find it better readable. But that might be just me. Is there a convention somewhere? I couldn't find anything in the code style guide.

spek

assertchris: Thanks for the work! :)

This kind of alignment isn't part of the current code style guidelines; and when I last tried to align values in this way, @tractorcow told me off.

In his book, Clean Code, Robert Martin talks a bunch about the relationship between alignment and readability. There's lots of good stuff in there, but he concludes that most of the time it doesn't help. There are exceptions, but I don't think aligning multiple array properties (like this) is one of them. If each array's keys and values were aligned then perhaps it could help...

assertchris:
Also, when you're ready, would you mind squashing these into a single commit?

code/extensions/ScheduledExecutionExtension.php
spekulatius: Good to know there isn't a fix definition. I used to do un-aligned definition and switched later on because I felt its better to read. But that might be just me.

I had a quick look at "Clean Code" and going to give it a shot. It sounds like a good book to read.

spekulatius:
I squashed them into one commit @assertchris . Please have a look before you merge it.

assertchris:
Looks good, thanks! :)

spekulatius:
Sweet, thanks for merging this in :)

PR #65, opened 2015-09-20
Codestyle adjustments and adding some phpdoc blocks

spekulatius:


code/jobs/ScheduledExecutionJob.php
```diff
@@ -9,7 +9,10 @@
  * @license BSD License http://silverstripe.org/bsd-license/
  */
 class ScheduledExecutionJob extends AbstractQueuedJob {
-
+	/**
+	 * @param DataObject $dataObject
+	 * @param integer $timesExecuted
```
assertchris: `int` please :)

spekulatius: yesss, of course.

spekulatius: I have fixed this @assertchris . Thanks for looking at this :)

PR #76, opened 2015-11-19
Added standard license

helpfulrobot:
Hello!

I'm helpful robot. Isn't the weather acceptable in [subject home town]? I was just walking my [subject favourite pet] and thinking about how I could help you make your license file better. Do you like what I've come up with?

Have a nice day! [shurp shurp]


tractorcow:
I think it might be best to leave this one alone.

@nyeholt do you agree with me closing this without merge?

PR #83, opened 2016-02-02
BUG Fix error logging passing invalid errcontext

dhensby:
This is a squash and rebase of #17


dhensby:
ping @tractorcow

tractorcow:
Oh right, good fix. :)

PR #87, opened 2016-03-07
Remove CMS as composer requirement

swilsonalfa:
After some local testing, it seems that silverstripe-cms doesn't really need to be a requirement for this project.


tractorcow:
Thanks. :D

PR #88, opened 2016-03-11
Fix reference to queued job admin email

swilsonalfa:
The documentation suggests that the correct config variable for error emails is queued_job_admin_email, however, the code in the constructor of the QueuedJobService is set to queued_jobs_admin_email/


tractorcow:
Nice fix, thanks!

PR #90, opened 2016-04-03
BUG Remove invalid backtraces from errcontext

tractorcow:
Fixes https://github.com/silverstripe-labs/silverstripe-fulltextsearch/issues/111

This causes errors because SS_Backtrace cannot parse these arrays as being valid backtraces.


tractorcow:
Thanks @nyeholt !

PR #91, opened 2016-04-07
fix long job cron example in readme

sheadawson:
Looking at ProcessJobQueueTask.php - getQueue(), queue should be one of immediate, queued or large


tractorcow:
Thank you!

PR #95, opened 2016-06-05
FIX CheckJobHealth only checks the health of jobs in the current queue

### LINKED ISSUE #94: checkJobHealth doesn't distinguish between queues
It appears that `QueuedJobService::checkJobHealth()` doesn't take into account the current queue and therefore how long a job should be running before determining if it's stalled.

In the documented setup you'd have a cron running by the minute and by the quarter hour to run appropriate queues, however each of these crons will check for any task whatsoever that is "stalled" and will attempt to restart it.

This means that the "short" queue will try to run "long" jobs that appear to be "stalled" and the long task will be marked as paused quite quickly if progress isn't made fast enough.

This results in several processes running the same job and whatever side effects come with that.

dhensby:
Fixes #94

### DIFF
--- a/code/services/QueuedJobService.php
+++ b/code/services/QueuedJobService.php
@@ -270,17 +270,20 @@ class QueuedJobService {
 	 * This typically happens when a PHP fatal error is thrown, which can't be picked up by the error
 	 * handler or exception checker; in this case, we detect these stalled jobs later and fix (try) to
 	 * fix them
+     *
+     * @param int $queue The queue to check against
 	 */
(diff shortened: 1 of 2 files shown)


dhensby:
What's going on with the tests :(

tractorcow:
I guess a regression from https://github.com/silverstripe/silverstripe-framework/commit/4e6d0b39823cb239ca21e1412b1105db5ca68039?

tractorcow:
Someone broke 3.1 branches too. :P

PR #101, opened 2016-09-16
Add daily CleanupJob to remove stale DB entries

### LINKED ISSUE #44: Clear jobs from the db
So as not to clog up the db there should be a way to clear jobs from the db after a set interval, or after x completions. Use case being I have a job that runs every minute, over the course of a day thats 1440 new records.

andrewandante:
Fixes #44 

Introduces a CleanupJob that is off by default to ensure an upgrade wouldn't start destroying DB entries unexpectedly.

Cleans out anything in the DB with JobStatus "Complete" or "Broken" more than 30 days old in its default configuration. Can be tweaked to save n records, then clean out anything matching a valid status above that.

Tweaked the end of `runJob()` to ensure there is one CleanupJob in the queue, and if not, schedule on in for a day from then. Should mean that it runs once a day.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ developers set these processes to be executed in the future.
 The module comes with
 
 * A section in the CMS for viewing a list of currently running jobs or scheduled jobs.
-* An abstract skeleton class for defining your own jobs
+* An abstract skeleton class for defining your own jobs.
 * A task that is executed as a cronjob for collecting and executing jobs.
+* A pre-configured job to cleanup the QueuedJobDescriptor database table.
 
 ## Quick Usage Overview
 
@@ -120,6 +121,36 @@ inotifywait and then call the ProcessJobQueueTask when a new job is ready to run
 Note - if you do NOT have this running, make sure to set `QueuedJobService::$use_shutdown_function = true;`
 so that immediate mode jobs don't stall. By setting this to true, immediate jobs will be executed after
 the request finishes as the php script ends. 
+
+## Configuring the CleanupJob
+
+By default the CleanupJob is disabled. To enable it, set the following in your YML:
+
+```yaml
+CleanupJob:
+  is_enabled: true
+```
+This will ensure that the CleanupJob is run once a day.
+
+You can configure this job to clean up based on the number of jobs, or the age of the jobs. This is
+configured with the `cleanup_method` setting - current valid values are "age" (default)  and "number".
+Each of these methods will have a value associated with it - this is an integer, set with `cleanup_value`.
+For "age", this will be converted into days; for "number", it is the minimum number of records to keep, sorted by LastEdited.
+The default value is 30, as we are expecting days.
+
+You can also determine which JobStatuses are allowed to be cleaned up. The default setting is to clean up "Broken" and "Complete" jobs. All other statuses can be configured with `cleanup_statuses`.
+
+The default configuration looks like this:
+
+```yaml
+CleanupJob:
+  is_enabled: false
+  cleanup_method: "age"
+  cleanup_value: 30
+  cleanup_statuses:
+    - Broken
+	- Complete
+``` 
 
 
 ## Troubleshooting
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -121,6 +122,36 @@ Note - if you do NOT have this running, make sure to set `QueuedJobService::$use
...
+```yaml
+CleanupJob:
+  is_enabled: true
+```
+This will ensure that the CleanupJob is run once a day.
+
+You can configure this job to clean up based on the number of jobs, or the age of the jobs. This is
+configured with the `cleanup_method` setting - current valid values are "age" (default)  and "number".
+Each of these methods will have a value associated with it - this is an integer, set with `cleanup_value`.
+For "age", this will be converted into days; for "number", it is the total number of records to keep.
```
dhensby: > it is the total number of records to keep.

Is it not the *minimum* number to keep?

code/jobs/CleanupJob.php
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @author Andrew Aitken-Fincham <andrew@silverstripe.com.au>
+ */
+class CleanupJob extends AbstractQueuedJob implements QueuedJob {
+
+	/**
+	 * How we will determine "stale"
+	 * Possible values: age, number
+	 * @config
+	 * @return string
```
dhensby: These should be `@var` not `@return`

dhensby (changes requested):
Looking good - just a few minor docs issues

README.md
andrewandante: True. Will update.

code/jobs/CleanupJob.php
andrewandante: Whoops.

dhensby (approved):

Firesphere:
Have you thought of adding tests?

andrewandante:
Do you mean in addition to `tests/CleanupJobTest.php`? Were you envisaging others?

Firesphere:
I'm blind. Or you secretly force-pushed with an adjusted date. I'm guessing the first.

andrewandante:
@nyeholt any movement or feedback on the above?

code/services/QueuedJobService.php
```diff
@@ -615,14 +615,20 @@ public function runJob($jobId) {
...
+		// okay let's reset our user if we've got an original
 		if ($runAsUser && $originalUser) {
 			Session::clear("loggedInAs");
 			if ($originalUser) {
 				Session::set("loggedInAs", $originalUser->ID);
 			}
 		}
 
+		// let's make sure there is a cleanupJob in the queue
+		if (Config::inst()->get('CleanupJob', 'is_enabled')) {
```
nyeholt: Could the job itself be responsible for re-adding itself to the queue? The initial creation can be done directly from the UI - my thinking being that if you want to _not_ have the cleanup running, you're needing to make server level configuration changes rather than just deleting the pending job from the UI.

andrewandante: It could - I did consider that behaviour. As it is already disabled by default, I'm not sure which I prefer. I think it ultimately comes down to who you want controlling whether or not the cleanup job runs - the CMS admin, or the dev. My gut says the dev, and that it should be controlled by config, but I guess ultimately it is up to you as module maintainer. Happy to change it over to be a self-restarting job if that's what you think is best.

nyeholt: If your deployment process is something that takes 10s of minutes across multiple distributed backend nodes, that requires a convoluted approval process to perform, I'd much prefer being able to simply remove it from the UI and not have it restart rather than go through that if I need to stop it from running.

andrewandante: Point taken. Moved the regeneration of the job to the job itself!

dhensby:
@nyeholt ping

nyeholt:
Sorry, didn't get the followup commit notification - all good!

Issue #99, opened 2016-08-12
Completed queued jobs in admin


Firesphere:
Currently, queued jobs "disappear" after they've ran. It would be nice to have a completed jobs section, to view the messages. As people aren't keeping an eye on their system 24/7.

tractorcow:
Yeah that might be nice. Maybe on another tab right? Kind of like how comments module splits "new", "spam", and "moderated" comments into tabs.

Firesphere:
Created a pull request for this. Implemented in a slightly different way because it felt more logical to rename the class to "FinishedJob", as it's not a description of a QueuedJob anymore.

robbieaverill:
For reference the PR is #102

PR #102, opened 2016-09-18
#99 Add finished jobs in a separate tab

Firesphere:
Also added the messages to the job view, by using a separate class.
The classname is updated before write when it's complete.

Updated test to fetch the FinishedJob instead of the QueuedJobDescriptor.

### DIFF
--- a/code/dataobjects/FinishedJob.php
+++ b/code/dataobjects/FinishedJob.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * Created by IntelliJ IDEA.
+ * User: simon
+ * Date: 18/09/16
+ * Time: 17:05
+ */
+class FinishedJob extends QueuedJobDescriptor
+{
+
+    public function canCreate($member = null)
+    {
+        return false;
+    }
+
+    public function canDelete($member = null)
+    {
+        return true;
+    }
+
+    public function canEdit($member = null)
+    {
+        return false;
+    }
+
+    public function canView($member = null)
+    {
+        return true;
+    }
+
+    public function getCMSFields() {
+        $fields = parent::getCMSFields();
+        $fields->addFieldToTab('Root.Main', LiteralField::create('Messages', $this->getMessages()));
+        return $fields;
+    }
+}
--- a/code/services/AbstractQueuedJob.php
+++ b/code/services/AbstractQueuedJob.php
@@ -35,14 +35,14 @@ abstract class AbstractQueuedJob implements QueuedJob {
 	 * @var boolean
 	 */
 	protected $isComplete = false;
-	
+
 	/**
 	 * Extensions can have a construct but don't have too.
 	 * Without a construct, it's impossible to create a job in the CMS
 	 * @var array params
 	 */
 	public function __construct($params = array()) {
-	    
+
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


nyeholt:
Is there any reason it couldn't just use a different filter mechanism on the main model admin view? Whereby the 'default' state of the filtering would be to only display completed jobs, but allows someone to untick that?

Then update the QueuedJobDescriptor's can* methods to account for the Complete status.

Firesphere:
My thought was, to move them to a separate Class, would be a cleaner way to handle it.

code/dataobjects/FinishedJob.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        return true;
+    }
+
+    public function canEdit($member = null)
+    {
+        return false;
+    }
+
+    public function canView($member = null)
```
nyeholt: Can this be restricted to admin users only?

code/services/AbstractQueuedJob.php
```diff
@@ -136,7 +136,7 @@ public function jobFinished() {
 	 * Called when the job is determined to be 'complete'
 	 */
 	public function afterComplete() {
-
+        $this->ClassName = 'FinishedJobs';
```
nyeholt: FinishedJob~~s~~

nyeholt:
My concern though is that it's adding additional complexity in the post-complete phase that's not really necessary for the usecase required. If there are additional things that could make use of the separate object, that might make it a useful addition

code/dataobjects/FinishedJob.php
Firesphere: Sure, note added to my list of things-to-do :)

code/services/AbstractQueuedJob.php
Firesphere: Good spot!

Firesphere:
I would think, having the classname changed, would also make a cleanup job, as per #101 easier. Plus, the added benefit is that the job runner only has to scan the QueuedJobs and thereby not having to run through possibly late finished jobs.
The latter was an issue I ran into, when having jobs running that take longer than expected, and "restart" the job, multiple times due to that. By changing it's classname, the first run finishing late, would change the classname to finished and by then, that issue didn't occur anymore for any subsequent automated crontasks trying to run it.

dhensby:
> I would think, having the classname changed, would also make a cleanup job, as per #101 easier. Plus, the added benefit is that the job runner only has to scan the QueuedJobs and thereby not having to run through possibly late finished jobs.

I actually think it makes things harder. At the moment that PR would allow devs to dictate which jobs can be cleaned by status. Now it'd only be able to clean up ones that completed and changed the class correctly - or it'd need refactoring to clean up multiple object types.

nyeholt:
> when having jobs running that take longer than expected, and "restart" the job, multiple times due to 
> that. By changing it's classname, the first run finishing late, would change the classname to finished and 
> by then, that issue didn't occur anymore for any subsequent automated crontasks trying to run it.

So the classname changed before the job is actually fully complete? Because that could be very problematic - either way, I don't really like changing the classname as part of the execution. I'd really prefer a filter option on the side - and if there's a real need to have them as a separate job, that that happens as a post-fact archiving step.

robbieaverill:
It doesn't sound like there's an agreement from the maintainers that this feature fits the project, and since it's pretty stale now I'm going to close it. Feel free to reopen if you'd like to revisit it again! Thanks for contributing =)

Issue #114, opened 2017-02-12
"Start Job At" date is interpreted in American format when entered in European format.


sapiensfio:
If a user has their DateFormat setting set to "dd/MM/yyyy" and creates a new job with the "Start Job At" field set to "01/04/2017" (for example) then the StartAfter field for this job is saved as "2017-01-04" when it should be "2017-04-01". 

This is because PHP interprets dates separated by a '/' as being in American format.

PR #115, opened 2017-02-12
Solved issue #114 where a job's StartAfter date was always being save…

sapiensfio:
…d in American format irrespective of the user's chosen DateFormat setting.


nyeholt:
I'd always just entered dates in YYYY-mm-dd format; never even occurred to me that people would do it other ways :) . I was trying to come up with a more elegant solution, but this looks to handle it as directly as possible.

PR #119, opened 2017-04-01
Making inclusion-duration of finished jobs configurable

micschk:
PR against 3.1+ branch


micschk:
Would this be OK to be merged & tagged into the current 3.x branch?

PR #120, opened 2017-04-02
feat(devjobsadmin): Add dev/jobs web/CLI interface for Queued Jobs

silbinarywolf:
Added the functionality as per my issue here:
https://github.com/silverstripe-australia/silverstripe-queuedjobs/issues/117

Screenshot:
screen1
screen2

### DIFF
--- a/code/controllers/QueuedJobsDevAdmin.php
+++ b/code/controllers/QueuedJobsDevAdmin.php
@@ -1,0 +1,239 @@
+<?php
+
+class QueuedJobsDevAdmin extends Controller {
+	private static $url_handlers = array(
+		'' => 'index',
+		'$JobClassName!/$ID' => 'runjob'
+	);
+
+	private static $allowed_actions = array(
+		'index',
+		'runjob',
+	);
+
+	/**
+	 * @var array
+	 */
+	private static $dependencies = array(
+		'jobQueue' => '%$QueuedJobService',
+	);
+
+	public function init() {
+		parent::init();
+
+		$isRunningTests = (class_exists('SapphireTest', false) && SapphireTest::is_running_test());
+		$canAccess = (
+			Director::isDev()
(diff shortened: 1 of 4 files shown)


SilbinaryWolf:
@nyeholt Feedback on this please?

code/controllers/QueuedJobsDevAdmin.php
```diff
@@ -0,0 +1,276 @@
...
+				if ($job) {
+					if (Director::is_cli()) {
+						echo "Starting existing job #$job->ID $job->Implementation \n";
+					}
+					return $this->executeJobAndRender($job->ID, 
+					'Successfully ran existing job #{$ID} "{$Class}"',
+					'Broken existing job #{$ID} "{$Class}"');
+				}
+				// If no items exist, create one with default parameters and execute.
+				$job = method_exists($class, 'create') ? $class::create() : new $class;
```
nyeholt: Is it possible to provide a list of constructor args to be used in this circumstance? Many jobs types expect constructor params for initialisation, which this doesn't cater for. 

Also, can you check that $class implements QueuedJob before creating an instance? Rather than an arbitrary URL-supplied class being created here.

silbinarywolf:
I still want to finish this PR. Just need to rebase on master I suppose.

robbieaverill:
@silbinarywolf thanks for contributing, and sorry we didn't get this over the line.

My suggestions from here:

* Rebase this work and remove any overlap with Sam's work in #151. It looks like this PR still adds new value, so would be cool to get it in.
* Switch the branch to a major branch e.g. 3 so we can release this as a new feature rather than a patch fix.
* Check for SilverStripe 4 compatibility. We're now in "critical bug fix and security only" mode for SilverStripe 3 support.

Cheers, and hope to see this come up again!

silbinarywolf:
@robbieaverill It wont! I noticed Sam's addition recently and I feel this is good enough :)

Issue #122, opened 2017-04-10
Update asyncphp/doorman to ^3.0


robbieaverill:
Just logging this here to help alleviate a composer dependency chain breaking with the latest SS4 builds:

* silverstripe/framework requires **nikic/php-parser ^2 || ^3**
* silverstripe/activedirectory requires *silverstripe/queuedjobs* ^3.0@dev
* silverstripe/queuedjobs requires asyncphp/doorman ~1.2
* asyncphp/doorman requires jeremeamia/superclosure ^1.0
* jeremeamia/superclosure requires **nikic/php-parser ~0.9**

This means you can't update the framework and CMS if you've got queuedjobs installed.

The latest version of asyncphp/doorman (3.0.0) requires superclosure ^2.0 which in its latest version (2.3.0) has a much looser constraint for php-parser: ^1.2 | ^2.0 | ^3.0, which would be compatible with the framework.

I'm not familiar with doorman, but can have a crack at updating this requirement and will PR a fix.

sminnee:
If doorman can use superclosure 2.3.0 we should be okay.

sminnee:
FYI @robbieaverill doorman was built by @assertchris

robbieaverill:
Closing - done in #125

PR #123, opened 2017-04-10
Update doorman to ^3.0

robbieaverill:
I haven't tested this thoroughly yet, but have made the required API change to the DoormanProcessManager since ProcessManager in doorman is now a final class.

Issue: #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ sudo: false
 language: php
 
 php:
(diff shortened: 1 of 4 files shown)


robbieaverill:
Looks like there's some other SS4 related updates that need to be made.

robbieaverill:
Replaced by #125

PR #124, opened 2017-04-19
FIX: Update doorman to 3.0 in composer.json

brettt89:
Update asyncphp/doorman to 3.0 to be compatible with SilverStripe 4.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "silverstripe/framework": "^4@dev",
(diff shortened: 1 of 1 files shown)


nyeholt:
Hi Brett - Looks like all the travis tests are exploding - I have a feeling this isn't directly related to the doorman update, but is there a potential version mismatch there somewhere?

brettt89:
Yeah it looks like the checks are failing on ModelAdmin which I think is unrelated to these changes.

PR #125, opened 2017-04-21
SilverStripe 4.0 Compatability

brettt89:
Update module to be compatible with SilverStripe 4.0


src/Services/QueuedJobService.php
```diff
@@ -680,7 +681,7 @@ public function runJob($jobId)
                 }
             } catch (Exception $e) {
                 // okay, we'll just catch this exception for now
-                $this->getLogger()->error($e->getMessage());
+                $this->getLogger()->debug($e->getMessage());
```
sminnee: Why have you changed the level of these messages?

brettt89: I had done so after reading https://docs.silverstripe.org/en/4/developer_guides/debugging/error_handling/

However I misread it as "debug and warn" should be through logger and "user_error" for errors.

brettt89:
Have been running through some tests manually and have found a couple of bugs that don't look to be covered by tests. Working through those now.

brettt89:
Date fields weren't updated. Updating these and adding tests for checking correct dates.

edlinklater:
Fixes #126

brettt89:
Fixes #122

tests/QueuedJobsTest/TestQJService.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace SilverStripe\QueuedJobs\Tests;
+
+use SilverStripe\QueuedJobs\Services\QueuedJobService;
+
+// stub class to be able to call init from an external context
+class TestQJService extends QueuedJobService
```
robbieaverill: These stubs should probably implement `SilverStripe\Dev\TestOnly` (I know they didn't before)

tests/ScheduledExecutionTest.php
```diff
@@ -118,9 +125,10 @@ public function testScheduledExecutionInterval()
     }
 }
 
-
-class TestScheduledDataObject extends DataObject implements TestOnly
+class TestScheduledDataObject extends DataObject
```
robbieaverill: Add test only back

brettt89: If I add "TestOnly" to these objects, the tests fail as SapphireTest ignores buildings the database for these objects.

https://github.com/silverstripe/silverstripe-framework/blob/master/src/Dev/SapphireTest.php#L1173

And the DB table is required due to the write occuring in the test
https://github.com/brettt89/silverstripe-queuedjobs/blob/compat-4.0/tests/ScheduledExecutionTest.php#L33

tests/QueuedJobsTest/TestQJService.php
brettt89: If I add "TestOnly" to these objects, the tests fail as SapphireTest ignores buildings the database for these objects.

https://github.com/silverstripe/silverstripe-framework/blob/master/src/Dev/SapphireTest.php#L1173

tests/ScheduledExecutionTest.php
robbieaverill: Sounds like a bug to me!

tests/ScheduledExecutionTest.php
```diff
@@ -11,11 +11,18 @@
...
+     *
+     * @var bool
+     */
+    protected $usesDatabase = true;
+
+    /**
      * {@inheritDoc}
      * @var array
      */
     protected $extraDataObjects = array(
```
sminnee: this should be `protected static $extra_dataobjects`

brettt89:
@sminnee @robbieaverill Have updated tests to correctly implement TestOnly.

Warnings in tests are due to expected errors thrown by QueuedJobs.

sminnee:
OK those warnings block merge. We'll need to think of a better way of handling these.

Perhaps the tests need to mock in a logger than can pick up these issues and validate that they have appeared?

robbieaverill:
Yeah, you can push a NullHandler from monolog to the logger from the setUp method so it won't default to logging these to the terminal

brettt89:
@sminnee @robbieaverill Have pushed a NullHandler onto the ErrorHandler using Injector. PHP is not reporting them as "Warnings", however the messages are still appearing in the output. Is that going to be an issue?

sminnee:
It makes it look broken, which inhibits all future dev, so yes I think we want to sort it out.

sminnee:
Are you sure that the messages are, in fact, benign?

src/QJUtils.php
```diff
@@ -5,6 +5,7 @@
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\DB;
+use Monolog\Logger;
```
robbieaverill: Is this used in the class?

.travis.yml
```diff
@@ -5,7 +5,6 @@ sudo: false
 language: php
 
 php:
-  - 5.5
```
robbieaverill: May as well add 7.1 while you're there

brettt89: Added in commit 2af7805bbc37d459e0a829cda2867cf0aed7eeac

brettt89:
@sminnee Have updated QueuedJobsTest to catch errors and check that expected errors exist after specific tests. Have confirmed that errors are expected as it is testing how the system reacts after failure.

@robbieaverill Have removed reference to Monologger/Logger and added tests for 7.1

nyeholt:
@sminnee did Brett's last update resolve the issue you had?

sminnee:
Yeah I'm happy with this now — no gunk in the test output.

I'd note that we probably don't need so many build configs and that the following should be sufficient:

 * PHP 5.6 + MYSQL
 * PHP 7.0 + PGSQL
 * PHP 7.1 + MYSQL

There's a finite number of concurrent builds across all `github.com/silverstripe-australia` runs so it's good to keep your job counts in check. ;-)

robbieaverill:
Assuming this PR fixes #122 (it's noted above as doing so, and the builds pass so I assume that's true) then it replaces #123 too (also replaced by #124). I've closed #123.

PR #128, opened 2017-05-10
fix(JobErrorHandler): Output error messages in console again

SilbinaryWolf:
**WARNING: This fix seems to mess with composer.json in unintended ways, what should I be doing here?**

fix(JobErrorHandler): Fix bug where deprecation / variable set in if-statement would always cause zero outputting of errors to console. Modified logic to align with Core silverstripe functions.

Culprit:
https://github.com/silverstripe-australia/silverstripe-queuedjobs/blob/master/src/Services/QueuedJobService.php#L969

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ By default the CleanupJob is disabled. To enable it, set the following in your Y
 CleanupJob:
   is_enabled: true
 ```
-This will ensure that the CleanupJob is run once a day.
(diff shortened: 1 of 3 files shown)


nyeholt:
Can this be rebased?

SilbinaryWolf:
Against what branch for 3.x stuff?

PR #132, opened 2017-07-03
Vendor update

jonom:
Got a warning when doing a composer update, this should fix it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 	"require":
 	{
 		"silverstripe/framework": "~3.1",
-		"silverstripe/multivaluefield": "~2.0",
+		"symbiote/silverstripe-multivaluefield": "~2.0",
 		"asyncphp/doorman": "~1.2"
 	},
     "extra": {


jonom:
Yeah it's just a suggestion:

> Package silverstripe/multivaluefield is abandoned, you should avoid using it. Use symbiote/silverstripe-multivaluefield instead.

But I guess that changing the vendor name on all the (previously) silverstripe-australia modules could lead to two potential headaches:

1. I'm assuming that since the `symbiote/silverstripe-multivaluefield` package is abandoned on Packagist, Packagist won't pick up any new tags pushed to the repo, so anyone requiring `silverstripe/multivaluefield` will be version locked and not benefit from any future releases.
2. Composer is happy to try to install both the `symbiote` and `silverstripe` (or `silverstripe-australia)` versions of a module side by side. This can happen if the project or one module requires the `symbiote` version and another module requires the `silverstripe` version, or vice versa. I assume both installations happen but the second instance overwrites the first. Currently this isn't an issue as they're installed in the same folder and the code matches, but down the track I could imagine this could lead to a franken-module, or both modules could be installed side by side leading to conflicts.

So I'm thinking that for the smoothest outcome, it would be best for any modules that have symbiote dependencies to update the vendor to `symbiote` as soon as possible?

nglasl:
1. That is correct. The previous packagist entry should no longer receive updates, so to benefit from future releases, the new `symbiote` packagist entry will need to be used. This encourages a swap to the new vendor name when looking for the most up to date version :)
2. This is something that actually popped up during the transition, where one module would end up at **queuedjobs**, and another at **silverstripe-queuedjobs**. This was due to the updated composer name (now having **silverstripe-** at the start), and was resolved by ensuring an `installer-name` existed. The problem no longer exists (when using the two latest major versions). If composer is attempting to pull down two different versions over each other, it will actually throw a dependency resolution error, so that shouldn't be a worry.

So yes, changing over any dependencies to `symbiote` as soon as possible is the ideal scenario, and this has already been done for our modules on each master branch. The next step is to work towards having this change propagated among the community :)

PR #135, opened 2017-07-05
Fix(defaultJobs) config now loads correctly. Add SS_Log of missing jo…

stephenmcm:
…b. Change admine email queued_job_admin_email

### DIFF
--- a/code/services/QueuedJobService.php
+++ b/code/services/QueuedJobService.php
@@ -88,7 +88,7 @@ class QueuedJobService {
 	 * Config controlled list of default/required jobs
 	 * @var Array
 	 */
-	public $defaultJobs;
+	public $defaultJobs = array();
 
 	/**
 	 * Register our shutdown handler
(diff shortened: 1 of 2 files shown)


code/services/QueuedJobService.php
```diff
@@ -325,6 +325,9 @@ public function checkJobHealth($queue = null) {
 	 */
 	public function checkdefaultJobs($queue = null) {
 		$queue = $queue ?: QueuedJob::QUEUED;
+
+		$this->defaultJobs = !empty($this->defaultJobs) ? $this->defaultJobs : Config::inst()->get(__CLASS__, 'defaultJobs');
```
nyeholt: Why isn't this using Injector configuration ? ie

```
Injector:
  QueuedJobService:
    defaultJobs: 
```

stephenmcm: Missing the `properties` level nest there:
```
Injector:
  QueuedJobService:
    properties:
      defaultJobs: 
```

I missed that while testing so had had the config sitting outside the Injector config array. I've updated the readme to add it and remove the unnecessary check.

nyeholt: The point is that it should never be calling `Config::inst()->get(__CLASS__, 'defaultJobs');` - defaultJobs is declared as a member variable (which should be injector set) rather than a private static (Config:: managed).

PR #138, opened 2017-08-29
loosened version requirements of silverstripe/admin package

bendubuisson:


dhensby:
OK - master is currently broken, so this doesn't introduce any regressions.

I can't merge failing builds as I'm not an admin so we have to wait for someone who can...

bendubuisson:
thanks @nyeholt ! Could we get a release please? I believe it is `4.0.5` ? Cheers

PR #141, opened 2017-09-24
FIX Update "Create new job" button to use bootstrap and escape HTML in messages in GridField

robbieaverill:
Before:

image

After:

image

### DIFF
--- a/src/Controllers/QueuedJobsAdmin.php
+++ b/src/Controllers/QueuedJobsAdmin.php
@@ -10,12 +10,17 @@ use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FormAction;
 use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\Forms\GridField\GridFieldAddNewButton;
 use SilverStripe\Forms\GridField\GridFieldConfig_RecordEditor;
+use SilverStripe\Forms\GridField\GridFieldDataColumns;
 use SilverStripe\Forms\TextareaField;
 use SilverStripe\ORM\DataList;
-use Symbiote\QueuedJobs\Forms\GridFieldQueuedJobExecute;
-use Symbiote\QueuedJobs\Services\QueuedJob;
 use SilverStripe\Security\Permission;
+use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
+use Symbiote\QueuedJobs\Forms\GridFieldQueuedJobExecute;
+use Symbiote\QueuedJobs\Services\AbstractQueuedJob;
+use Symbiote\QueuedJobs\Services\QueuedJob;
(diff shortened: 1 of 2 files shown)


robbieaverill:
Tests are fixed in #139

nyeholt:
A quick question on convention here... to better support SS3 => SS4 migrations, I had been thinking that a simpler approach was to define

```
SilverStripe\Core\Injector\Injector:
  QueuedJobService:
    class: \Symbiote\QueuedJobs\Services\QueuedJobService
    # etc
```

That would mean existing `singleton('QueuedJobService')` calls would continue to work. Alternatively something that re-registers the namespaced label into the injector at the 'short' name would work, but feels a little cumbersome. Thoughts? 

(Though that discussion doesn't block this as it stands)

robbieaverill:
@nyeholt yeah I'd wondered whether you wanted to keep injector aliases for various things. I personally don't think you need it for services like this, but it might be useful for user definitions of config in the advanced workflow module for example. Up to you :)

nyeholt:
I actually prefer the alias style of doing things, as otherwise your configuration is binding to a specific namespace wherever you reference that object, which is counter to one of the goals of using a DI. Probably a broader topic to be raised on the dev mailing list

robbieaverill:
For some context- the way this is generally done in SS4 core is that injector services are bound to an interface name, which is then configured in YAML to a specific class (and can be adjusted without breaking APIs, because you expect an interface instance).

That said, in some cases it doesn't really seem necessary to add an interface just to return a single class.

sminnee:
> I actually prefer the alias style of doing things, as otherwise your configuration is binding to a specific namespace wherever you reference that object, which is counter to one of the goals of using a DI. Probably a broader topic to be raised on the dev mailing list

In core code we've tried to take the approach of binding to the FQCN of a relevant interface. It still has a namespace but it's a namespace related to the problem domain rather than the implementation.

Ingo and I had arguments about aliases vs FQCNs at the time:

 * Ingo advocated aliases for brevity's sake (particularly when hand-writing in YML)
 * I advocated FQCNs because otherwise cross-module namespace collisions would still be a risk, which is the whole point of namespaces, and you'd end up with 2 names for everything, which doesn't seem less confusing.

I won the argument at the time.

Like you say, probably something for the dev list, but would be keen to get your views.

sminnee:
> A quick question on convention here... to better support SS3 => SS4 migrations, I had been thinking that a simpler approach was to define

If you create a file calle `.upgrade.yml` in the module with this text:

```
mappings:
  QueuedJobService: Symbiote\QueuedJobs\Services\QueuedJobService
```

Then it should be taken care of for you? It performs exact-match replacements of "QueuedJobService" string literals too, so the singleton calls will be sorted out.

The alias approach might introduce weird edge-case bugs since if code expected to grab the same service in 2 different places but actually had different service instances.

PR #146, opened 2017-11-02
Add scheduledJob to ImmediateQueueHandler

dhensby:
This API is relied upon in `QueuedJobHandler` and causes fatal errors if you try to use the immediatequeuehandler.


robbieaverill:
PHP 5.3 failure in Travis is unrelated

dhensby:
> I feel like there should be a QueueHandlerInterface to prevent this sort of thing from happening

indeed - but adding interfaces can be considered a breaking change and so needs to go into a major release line :(

robbieaverill:
Logged for future reference as #149

PR #151, opened 2017-11-10
NEW: Allow queueing of build tasks

### LINKED ISSUE #150: Feature: Ability to queue build tasks
The queuedjobs module should come out of the box with a facility for executing build tasks as a job. This would mean that sites that have a number of build tasks already written can easy shift to running these in a back-end queue.

My current use-case is to limit the need to SSH on a box to perform server-admin tasks, which often take the form of long-running build tasks.

Core features:

 * Create a RunBuildTaskJob job type that can execute any build task
 * Replace the default /dev/tasks UI to provide one that can enqueue a RunBuildTaskJob instead of running the job immediately

Optional features:

 * Provide an email address where the output of the buildtask will be sent.

sminnee:
This adds a couple of items:

 * A job type RunBuildTaskJob, that will run a single build task
 * A QueuedTaskRunner, that overrides the default dev/tasks admin in non-CLI mode, to allow queuing of build tasks by default

The result is that when queued jobs is installed, web management of build tasks will by default happen on background job queues. This will simplify the deployment of SilverStripe sites to servers where the dev team don’t have SSH access.

Fixes #150

### DIFF
--- a/code/jobs/RunBuildTaskJob.php
+++ b/code/jobs/RunBuildTaskJob.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * A job used to delete a data object. Typically used for deletes that need to happen on
+ * a schedule, or where the delete may have some onflow affect that takes a while to
+ * finish the deletion.
+ *
+ * @author marcus@symbiote.com.au
+ * @license BSD License http://silverstripe.org/bsd-license/
+ */
+class RunBuildTaskJob extends AbstractQueuedJob {
+    /**
+     * @param DataObject $node
+     */
+    public function __construct($taskClass = null, $queryString = null) {
+        if ($taskClass) {
+            $this->TaskClass = $taskClass;
+        }
+
+        if ($queryString) {
+            $this->QueryString = $queryString;
+        }
+
+        $this->currentStep = 0;
+        $this->totalSteps = 1;
+    }
+
+    /**
+     * @param string (default: Object)
+     * @return DataObject
+     */
+    protected function getObject($name = 'Object') {
+        return DataObject::get_by_id($this->TargetClass, $this->TargetID);
+    }
+
+    /**
+     * @return string
+     */
+    public function getJobType() {
+        return QueuedJob::QUEUED;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTitle() {
+        $taskName = $this->QueryString ? ($this->TaskClass . '?' . $this->QueryString) : $this->TaskClass;
+        return _t('RunBuildTaskJob.JOB_TITLE', 'Run BuildTask {task}', array('task' => $taskName));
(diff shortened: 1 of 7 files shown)


sminnee:
OK it looks like the build failure is that the PHP 5.3 test run won't work on Travis default machines. I believe they're dropping 5.3 support entirely next year; should we remove it from the test run, or update the builds to run on "precise" rather than "trusty"?

nyeholt:
I'm happy to just remove 5.3 - it's not supported for SS4 right?

sminnee:
Yeah, SS4 is PHP 5.6+.
SS3.6+ supports 5.3 - 7.1.

nyeholt:
Happy to merge as-is, but do you have any idea off the top of your head as to why PHP5.5 is being a jerk (but not others, go figure) 

https://travis-ci.org/symbiote/silverstripe-queuedjobs/jobs/301742025

sminnee:
PS I should also point out that 73135e676e6e7154e6f35712c82831efa257ace7 and 74c13c628342f1b20b78a64ede1b5cbd881e2556 are completely different features that were also chucked in this branch. If you object to either of those, let me know and I'll untangle things.

sminnee:
> We could drop 3.1 support, or I could keep nudging that file until it works.

On second thoughts, that would be a major release bump, so I won't do that. ;-)

sminnee:
OK @nyeholt I've fixed the broken test and squashed so that each feature is a separate commit.

sminnee:
Hola – I'd be keen to get this merged so that we can use it on an updated addons.silverstripe.org.

code/jobs/RunBuildTaskJob.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+/**
+ * A job used to delete a data object. Typically used for deletes that need to happen on
+ * a schedule, or where the delete may have some onflow affect that takes a while to
+ * finish the deletion.
```
robbieaverill: Is this doc block correct? Looks like the job is for running rather than deleting

code/tasks/DeleteAllJobsTask.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param SS_HTTPRequest $request
+     */
+    public function run($request)
+    {
+        $confirm = $request->getVar('confirm');
+
+        $jobs = DataObject::get('QueuedJobDescriptor');
+
+        if (!$confirm) {
+            echo "Really delete " . $jobs->count() . " jobs? Please add ?confirm=1 to the URL to confirm.";
```
robbieaverill: 👍

nyeholt:
Hah, sorry mate, didn't realise you'd finished the push back in Nov. No problems with it as is!

PR #154, opened 2017-12-21
Fix(Service) ensure run as user is cleared at the end of each runJob

stephenmcm:


dhensby:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

stephenmcm:
Hi Dan,
As far as I can tell this is the right branch? While this does affect all versions of this module, I've only tested it against a SilverStripe 3.x. Looking at the branches here 2.10 appears to be the tip for SS 3.x.

robbieaverill:
Hey @stephenmcm, in this case I think @dhensby means you could use a branch name like `2.10/clear-user-after-job` rather than committing directly to the `2.10` branch =)

Issue #45, opened 2015-06-17
Minor change to GridField iconography


phptek:
I'm no UX expert but the icon used to represent a job to be paused , looks like a "No Entry" sign. My suggestion is to change the pause icon to look like a pause button does on a CD player, and the current "Reload" icon, while better, should use a "Play" icon - again as per the CD player analogy for consistency.

robbieaverill:
_Improvements_ made at #164, but probably not enough to close this issue.

robbieaverill:
@phptek have you had a chance to check out the new icons in SS4? Thoughts?

phptek:
@robbieaverill Yeah now looks awesome. Closing.

PR #164, opened 2018-01-26
FIX Correct field name in execute action handler and update icons to use admin icons

### LINKED ISSUE #159: Unable to see x3 GridField actions
PHP7.0 / framework+cms 4.0.x-dev / queuedjobs dev-master

* Login to the backend as an "ADMIN" user
* Navigate to the "jobs" admin
* Observe (or create) jobs listed in the GridField
* Observe no 'execute', 'pause' or 'resume' actions display in the "Actions" column

Only `GridFieldQueuedJobExecute::__construct()` is called, whereas placing simple debug statements in this classes' other implemented methods does not result in debug logic being called as I would expect. Hacking around in for example `GridFieldEditButton` in _its_ implemented methods, does (obviously) show that they're being called. Not sure what's happening here and cannot spend any more time on it today. (Maybe I'm missing something, it has been known...)

I tried going back as far as `3afac24f411` and buttons still not shown.

robbieaverill:
Two changes:

* Fix the field name for the QueuedJobDescriptor CMS field which was preventing the GridFieldQueuedJobExecute GridFieldComponent from being added in the Jobs admin area
* Change the icons for the job actions

The icons were old (gone) jQuery UI icon classes, and have been replaced with silverstripe-admin core icons. This helps towards #45, but doesn't fix it because the "pause" action is currently a circled cancel icon which doesn't really indicate "pause" very well. This fixes #159 though and is an improvement towards #45 without fixing it completely.

image


sachajudd:
Looks good @robbieaverill will keep in mind the "pause" action for a future core icon.

PR #169, opened 2018-03-15
Check that Security::getCurrentUser() returns not null before accessing property

chrispenny:
`QueuedJobService::queueJob()` is making the assumption that a `User` will always be provided or available through `Security::getCurrentUser()`, but `QueuedJobService::runJob()` does not require a `User` to be set, and a `User` might not need to be provided for Jobs queued through a cron.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -183,7 +183,11 @@ class QueuedJobService
         $jobDescriptor->Implementation = get_class($job);
         $jobDescriptor->StartAfter = $startAfter;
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
Ah so this has already been done with https://github.com/symbiote/silverstripe-queuedjobs/pull/167 in the 4.1.x release line. When we merge that up into master I'll probably use those changes to overwrite this PR

chrispenny:
Ooooh sorry! I did not see that one.

robbieaverill:
Neither =D

PR #171, opened 2018-03-15
Log QueuedJob errors with context provided.

chrispenny:
## TL;DR
`RaygunHandler` requires that context is provided to it. This can either be the `Exception` itself, or values for "line" and "file".

## Example trace for an `Exception`
`QueuedJobService`
Line `702` or `806`: An `Exception` is thrown and caught.
Line `704` or `808`: Logs the Exception message (*only* the message) to `Monolog\Logger::error()`.

`Monolog\Logger`
Line `614`: `error()` accepts `string $message` and `array $context` (please note, no `$context` was passed above, so this is an empty `array`)
Line `616`: `error()` calls `addRecord()` with error level, message, and context as it's params.
Line `322`: `addRecord()` creates `$record`:
```
$record = array(
    'message' => (string) $message,
    'context' => $context,
    'level' => $level,
    'level_name' => $levelName,
    'channel' => $this->name,
    'datetime' => $ts,
    'extra' => array(),
);
```
Please note that `context` is (from the trace) an empty `array`.
Line `337`: `addRecord()` calls `$handler->handle($record)`. This `$record` eventually makes it way to `Monolog\Handler\RaygunHandler::write()`.

`Monolog\Handler\RaygunHandler::write()`
```
$context = $record['context'];

if (isset($context['exception']) &&
    (
        $context['exception'] instanceof \Exception ||
        (PHP_VERSION_ID > 70000 && $context['exception'] instanceof \Throwable)
    )
) {
    $this->writeException(
        $record,
        $record['formatted']['tags'],
        $record['formatted']['custom_data'],
        $record['formatted']['timestamp']
    );
} elseif (isset($context['file']) && $context['line']) {
    $this->writeError(
        $record['formatted'],
        $record['formatted']['tags'],
        $record['formatted']['custom_data'],
        $record['formatted']['timestamp']
    );
} else {
    throw new \InvalidArgumentException('Invalid record given.');
}
```

`$record['context']` is an empty array, so neither of the above conditions are met. `RaygunHandler` then throws an `InvalidArgumentException`.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -361,7 +361,11 @@ class QueuedJobService
                         'errcontext' => array()
                     ),
                     true
-                )
+                ),
+                [
+                    'file' => __FILE__,
+                    'line' => __LINE__,
+                ]
             );
         }
     }
@@ -385,7 +389,13 @@ class QueuedJobService
             );
             foreach ($this->defaultJobs as $title => $jobConfig) {
                 if (!isset($jobConfig['filter']) || !isset($jobConfig['type'])) {
-                    $this->getLogger()->error("Default Job config: $title incorrectly set up. Please check the readme for examples");
+                    $this->getLogger()->error(
+                        "Default Job config: $title incorrectly set up. Please check the readme for examples",
+                        [
+                            'file' => __FILE__,
+                            'line' => __LINE__,
+                        ]
+                    );
                     continue;
                 }
                 $job = $activeJobs->filter(array_merge(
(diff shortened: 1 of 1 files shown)


dhensby:
Hmmm, the use of `__LINE__` is a bit misleading because it's going to report literally the line number of where the debug message is (well, actually of where the `__LINE__` is and so the context and the error message will contain different lines and they'll both not be the line the error was on...

chrispenny:
I'm open to other suggestions, but as it stands, this is still a lot more useful than an error from Raygun that simply says "Invalid record given".

Perhaps this is actually an issue for the `Monolog` module? Maybe we should be able to pass just a message without context?

robbieaverill:
_Possibly_ related? (Haven't read the details too much) https://github.com/silverstripe/silverstripe-framework/pull/7603 https://github.com/Seldaek/monolog/pull/1080

dhensby:
OK - can we drop the context completely?

chrispenny:
I don't think those issues are going to solve this particular problem with the `RaygunHandler`. It requires either an `Exception`, or line/file for context. If it doesn't get either, it throws an exception.

I think maybe the above issue is something that needs to be addressed on the `Monolog` module, so I will update this PR to just include the `Exception` context changes. I think that this much will still be useful.

I will ping when I've made those changes.

Cheers :)

dhensby:
Maybe `RaygunHandler` needs a patch?

sminnee:
Yeah – PSR-3 doesn't mandate the presence of a particular context array so I think RaygunHandler is violating expectations there.

That said, catching exceptions and logging the message without context isn't the best approach and so this patch is probably useful in addition to any changes to RaygunHandler.

dhensby:
ok, shall we take this in as-is?

sminnee:
I've raised a ticket re: RaygunHandler here https://github.com/graze/monolog-extensions/issues/20

PR #176, opened 2018-03-27
FIX: Clear the binary field

frankmullenger:
So that PHP is not used to interpret the sake bash script.

### DIFF
--- a/src/Tasks/Engines/DoormanRunner.php
+++ b/src/Tasks/Engines/DoormanRunner.php
@@ -3,6 +3,7 @@
 namespace Symbiote\QueuedJobs\Tasks\Engines;
 
 use AsyncPHP\Doorman\Rule;
+use AsyncPHP\Doorman\Manager\ProcessManager;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
 use Symbiote\QueuedJobs\Jobs\DoormanQueuedJobTask;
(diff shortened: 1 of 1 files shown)


tractorcow:
The task is run as `"{$binary} {$worker}"`, and since sake is ment to be run as a stand-alone process (not under php) this makes sense to me.

Without this prevent `"php vendor/bin/sake"` being incorrectly run.

@robbieaverill do you want to double check my thinking?

robbieaverill:
I don't have any context about this part of the code, but have just skim read the ProcessManager class and can say that:

* the binary property can be a string or null
* if it's null (would be before this PR changed it) it defaults to `$this->binary = PHP_BINDIR . "/php";`
* setting it to an empty string would prevent php from be used to interpret it
* it seems like a sneaky use of the API

Provided it does what the comment says I'm happy =)

robbieaverill:
@frankmullenger do you want to target a specific minor version branch so it'll become a patch release?

robbieaverill:
Oh, d'oh! This was caused by my commit at https://github.com/silverstripe-terraformers/silverstripe-queuedjobs/commit/35fa3c382dca145a2104fcb2f25a50c10f107373#diff-255a2898e87bc13a20569c932b00f920 🤦‍♂️ 

Previously: uses a PHP script
Now: uses a bash script

Options:

* merge this
* switch from using `vendor/bin/sake` to `vendor/silverstripe/framework/cli-script.php`

frankmullenger:
@robbieaverill ah, I see. Probably using the cli-script directly would be better/cleaner in this case. Do you want us to make those changes to this PR?

robbieaverill:
It might be a better option overall, eh? If you agree then yeah that'd be cool =)

tractorcow:
I think going back to the older behaviour is going to be safer in the long run;

robbieaverill:
@frankmullenger last check, are you happy with this going into master or do you want to target one of the minor release branches instead?

frankmullenger:
@robbieaverill happy with master thanks.

PR #178, opened 2018-04-04
NEW Allow ProcessManager log path to be configurable

### LINKED ISSUE #177: DoormanRunner suppresses ProcessManager errors
Currently cannot set the log path for the ProcessManager:
https://github.com/symbiote/silverstripe-queuedjobs/blob/master/src/Tasks/Engines/DoormanRunner.php#L53

The result is that errors thrown when exec runs the ProcessJobQueueChildTask go to /dev/null.

Preferably the log path can be set for the ProcessManager instance per project, perhaps a hook with the instance passed as an argument is the easiest option.

robbieaverill:
Resolves #177

### DIFF
--- a/src/Tasks/Engines/DoormanRunner.php
+++ b/src/Tasks/Engines/DoormanRunner.php
@@ -2,8 +2,8 @@
 
 namespace Symbiote\QueuedJobs\Tasks\Engines;
 
-use AsyncPHP\Doorman\Rule;
 use AsyncPHP\Doorman\Manager\ProcessManager;
+use SilverStripe\Core\Environment;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
 use Symbiote\QueuedJobs\Jobs\DoormanQueuedJobTask;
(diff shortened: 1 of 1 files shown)


tractorcow:
I think we should be using a PSR-3 logger, allowing this to be injected to doorman instead of a separate config string.

frankmullenger:
@tractorcow would that require a change to ProcessJobQueueChildTask and/or Doorman ProcessManager? 

At a minimum it will be useful for us to have the standard output and error streams sent somewhere useful:
https://github.com/asyncphp/doorman/blob/master/src/Manager/ProcessManager.php#L132

The problem being that it defaults to /dev/null:
https://github.com/asyncphp/doorman/blob/master/src/Manager/ProcessManager.php#L422

tractorcow:
ok, so it's just the bash shell logging, not something done at the PHP level.

I guess we could do this separately.

tractorcow:
@robbieaverill things that may referenced fixed paths (e.g. `/usr/var/log/doorman.log`) should probably be env-var, not config. Can you add support for an `SS_DOORMAN_LOGPATH` ?

platfrom-specific config is a bit harder, hence my concern. :)

frankmullenger:
Ah right, good point. Yes an environment variable would be preferred thanks.

robbieaverill:
Fair call! Updating

PR #180, opened 2018-04-06
Replace 'clipboard-pencil' font icon with 'checklist' font icon

### LINKED ISSUE #162: Update icon or icon colour to match other SS4 icons in the CMS
screen shot 2018-01-26 at 9 06 47 am
I suggest to use a font icon from core, perhaps the clipboard pencil one?
screen shot 2018-01-26 at 9 24 18 am
@sachajudd @clarkepaul

raissanorth:
Based on the emoji tally in the [related issue](https://github.com/symbiote/silverstripe-queuedjobs/issues/162), which I interpret to make the checklist font icon the crowd's favourite, it is used as the menu icon. This resolves #162 

It looks like that:
image

It works well with the other menu icons
image

The font icon was introduced to the admin module in https://github.com/silverstripe/silverstripe-admin/commit/d5d022ec752a38c7e931694ea6fa660f2fae85c4, so an update to the admin module version dependency might be required.

@clarkepaul

### DIFF
--- a/src/Controllers/QueuedJobsAdmin.php
+++ b/src/Controllers/QueuedJobsAdmin.php
@@ -43,7 +43,7 @@ class QueuedJobsAdmin extends ModelAdmin
     /**
      * @var string
      */
-    private static $menu_icon_class = 'font-icon-clipboard-pencil';
+    private static $menu_icon_class = 'font-icon-checklist';
 
     /**
      * @var array


robbieaverill:
Look good to you @clarkepaul?

clarkepaul:
Yup all good, pity its slightly smaller than the other icons but now that the icon is in we can swap it easily at any time (using the same classname).

PR #181, opened 2018-04-13
NEW: Ability to inject a different process manager class.

frankmullenger:
We would like to replace the manager with an instance of another class.


src/Tasks/Engines/DoormanRunner.php
```diff
@@ -47,7 +48,7 @@ public function runQueue($queue)
 
         // split jobs out into multiple tasks...
 
-        $manager = new ProcessManager();
+        $manager = Injector::inst()->create('AsyncPHP\Doorman\Manager\ProcessManager');
```
robbieaverill: This is all good, but I'd personally prefer to see `->create(ProcessManager::class)`

PR #185, opened 2018-05-06
FIX: Fixes #173 Check for existence of root object

### LINKED ISSUE #173: Inonsistent PublishItemsJob API
An instance of `Symbiote\QueuedJobs\Jobs\PublishItemsJob` can be instantiated with an optional `$rootNodeID`. But; the `getTitle()` method can still be called which is expecting `getRoot()` to return a `DataObject`, which it can't do if `$this->rootID` is null becuase no `$rootNodeID` is passed in construction.

The relevant part of the stack-trace where I have seen this in a SilverStripe 4 setup is as follows:

```
ORM/DataObject.php in get_by_id at line 3187
vendor/symbiote/silverstripe-queuedjobs/src/Jobs/PublishItemsJob.php in getRoot at line 34
vendor/symbiote/silverstripe-queuedjobs/src/Jobs/PublishItemsJob.php in getTitle at line 47
vendor/symbiote/silverstripe-queuedjobs/src/Services/QueuedJobService.php in queueJob at line 177
vendor/symbiote/silverstripe-queuedjobs/src/Controllers/QueuedJobsAdmin.php in createjob at line 190
```

This is just a bug-report for now. If I come up with a fix or suggestion, I'll of course update the ticket.

phptek:


dhensby:
Hmmm, those test failures... are they new or existing? They don't look like they should be caused by this change :/

robbieaverill:
This is the test that fails at 58/59 minutes past the hour =) restarted

PR #193, opened 2018-06-17
Fixed a case where original user was missing when unsetting a user.

mfendeksilverstripe:
Fixed a situation where unsetting a user would generate a PHP notice because we are trying to access object property of null. This is quite annoying as one notice per job is created.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -892,13 +892,15 @@ class QueuedJobService
             } else {
                 $_SESSION['loggedInAs'] = null;
             }
+
+            return;
         }
 
         // Okay let's reset our user.
         if (Controller::has_curr()) {
             Security::setCurrentUser($originalUser);
         } else {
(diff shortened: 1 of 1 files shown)


robbieaverill:
Do we need to target this PR at 4.0, 4.1, or 4.2? Or is master OK?

mfendeksilverstripe:
@robbieaverill This is actually super minor as I found that someone already fixed this issue in a later version. So my change doesn't do any functional change it only improves the code as it introduces an early exit. Up to you if you want to add it to master or not.

robbieaverill:
Ah OK. I can't see the same change in 4.1 or 4.2, perhaps we could switch the branch for this PR to 4.1? If you could do that and rebase the PR that'd be awesome

mfendeksilverstripe:
@robbieaverill Changed base to `4.2` as this is the branch where the issue was introduced. I couldn't find the instance of the related code in earlier versions.

PR #198, opened 2018-06-28
Ensure extension fields don't exist before adding them

wilr:
When applying this extension to a scaffolded dataobject, the fields already exist in the CMS so when this runs and tries to re-add SilverStripe will throw a warning indicating those fields already exist. This ensures that the fields have been removed first.

### DIFF
--- a/src/Extensions/ScheduledExecutionExtension.php
+++ b/src/Extensions/ScheduledExecutionExtension.php
@@ -54,10 +54,18 @@ class ScheduledExecutionExtension extends DataExtension
      */
     public function updateCMSFields(FieldList $fields)
     {
+        $fields->removeByName([
+            'ExecuteInterval',
+            'ExecuteEvery',
+            'ExecuteFree',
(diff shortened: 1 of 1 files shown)


src/Extensions/ScheduledExecutionExtension.php
```diff
@@ -54,6 +54,11 @@ class ScheduledExecutionExtension extends DataExtension
      */
     public function updateCMSFields(FieldList $fields)
     {
+        $fields->removeByName('ExecuteInterval');
+        $fields->removeByName('ExecuteEvery');
+        $fields->removeByName('ExecuteFree');
+        $fields->removeByName('FirstExecution');
```
robbieaverill: Minor thing, but you can pass an array to `removeByName`

wilr: Mind blown.

wilr:
Updated

PR #214, opened 2018-09-04
NEW Allowing disabling health checks - adding a task for it

### LINKED ISSUE #213: Having multiple runners will check job health too often
A job is considered broken if there have not been any "steps" completed since the last health check. This logic appears sound, but relies heavily on having a reasonable time period between health checks. 

Currently the health check is done when a runner starts running a queue. This is inherently flawed in a multi-worker environment, as 2 workers starting the same queue at the same time will immediately note all jobs are "stalled" - the first worker is happy, the second will think they're all stalled.

There are three potential fixes I can see
- Nominate workers that can perform health checks
- Move health checking to a separate cron - via a config setting
- Keep track of the last time a "health-check" was run and have a configurable time-period where another check is not performed

@frankmullenger What do you think would work best?

### PR:

#214

ScopeyNZ:
Resolves #213

This will do nothing by default. However you can now disable automated health checks with a config setting. There is also a new build task provided to do health checking - this can be set up as a cron.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,26 @@ Symbiote\QueuedJobs\Jobs\CleanupJob:
     - Complete
 ```
 
+## Health Checking
+
+Jobs track their execution in steps - as the job runs it increments the "steps" that have been run. Periodically jobs
+are checked to ensure they are healthy. This asserts the count of steps on a job is always increasing between health
+checks. By default health checks are performed when a worker picks starts running a queue.
+
+In a multi-worker environment this can cause issues when health checks are performed too frequently. You can disable the
+automatic health check with the following configuration:
+
+```yaml
+Symbiote\QueuedJobs\Services\QueuedJobService:
+  disable_health_check: true
+```
+
+In addition to the config setting there is a task that can be used with a cron to ensure that unhealthy jobs are
+detected:
+
+```
+*/5 * * * * php /path/to/silverstripe/vendor/bin/sake dev/tasks/CheckJobHealthTask
+```
 
 ## Troubleshooting
 
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -82,6 +82,18 @@ class QueuedJobService
     private static $time_limit = 0;
 
     /**
+     * Disable health checks that usually occur when a runner first picks up a queue. Note that completely disabling
+     * health checks could result in many jobs that are always marked as running - that will never be restarted. If
+     * this option is disabled you may alternatively use the build task
+     *
+     * @see \Symbiote\QueuedJobs\Tasks\CheckJobHealthTask
+     *
+     * @var bool
+     * @config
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -242,6 +242,26 @@ Symbiote\QueuedJobs\Jobs\CleanupJob:
...
+Jobs track their execution in steps - as the job runs it increments the "steps" that have been run. Periodically jobs
+are checked to ensure they are healthy. This asserts the count of steps on a job is always increasing between health
+checks. By default health checks are performed when a worker picks starts running a queue.
+
+In a multi-worker environment this can cause issues when health checks are performed too frequently. You can disable the
+automatic health check with the following configuration:
+
+```yaml
+Symbiote\QueuedJobs\Services\QueuedJobService:
+  disable_health_check: true
```
robbieaverill: I'd prefer this to be the opposite i.e. `health_check: false` where it's `true` by default

src/Services/QueuedJobService.php
```diff
@@ -1074,7 +1088,9 @@ public function getJobListFilter($type = null, $includeUpUntil = 0)
      */
     public function runQueue($queue)
     {
-        $this->checkJobHealth($queue);
+        if (!self::config()->disable_health_check) {
```
```diff
-         if (!self::config()->disable_health_check) {
+         if (!$this->config()->get('disable_health_check')) {
```

Also note my suggestion to flip this argument around so it'd be `if ($this->config()->get('health_check')) { ...`

src/Tasks/CheckJobHealthTask.php
```diff
@@ -0,0 +1,51 @@
...
+            'parameter or otherwise the "Queued" queue will be checked'
+        );
+    }
+
+    /**
+     * Implement this method in the task subclass to
+     * execute via the TaskRunner
+     *
+     * @param HTTPRequest $request
+     * @return
```
robbieaverill: Remove or specify void

src/Tasks/CheckJobHealthTask.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function run($request)
+    {
+        $queue = $request->requestVar('queue') ?: QueuedJob::QUEUED;
+
+        $stalledJobCount = $this->getService()->checkJobHealth($queue);
+
+        echo $stalledJobCount === 0 ? 'All jobs are healthy' : 'Detected and attempted restart on ' . $stalledJobCount .
+            ' stalled jobs';
```
robbieaverill: You localise the description, perhaps this should be localised as well?

README.md
ScopeyNZ: I felt like health checks are a given and you have to specifically disable them, but I don't really feel strongly.

NightJar: I also think about it that way @ScopeyNZ, and also do not feel strongly either way.

PR #217, opened 2018-09-17
email config

renskorswagen:


robbieaverill:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

robbieaverill:
Hi @renskorswagen, can you please explain why you're changing this? The idea is that you can configure a specific email address for queued job emails to go to - this PR kind of breaks that feature. Some background info would be nice =)

robbieaverill:
What are you expecting to happen with this code? Both sent and received from the admin email address?

renskorswagen:
@robbieaverill No, I want to use a different e-mail address for the sender and recipient. My admin_email is info@domain.com and my queued_job_admin_email is jobs@domain.com. I want all the emails send to the queued_job_admin_email and come from admin_email. For now there is no way to do that because of the different email settings on both lines.
https://github.com/symbiote/silverstripe-queuedjobs/blob/master/src/Services/QueuedJobService.php#L437
https://github.com/symbiote/silverstripe-queuedjobs/blob/master/src/Services/QueuedJobService.php#L505

renskorswagen:
@robbieaverill  You're right! Changed it.

robbieaverill:
Thanks for contributing!

PR #218, opened 2018-09-17
Change default job recreate log level to info

renskorswagen:


dhensby:
This seems to do more than the PR title implies. Please could you provide a description about what you're trying to achieve, please?

PR #224, opened 2018-11-21
CleanupJob query_limit

JoshuaCarter:
Adds the config option `query_limit` for the CleanupJob (and readme addition).

I would suggest a new release for this change.

### DIFF
--- a/code/jobs/CleanupJob.php
+++ b/code/jobs/CleanupJob.php
@@ -41,6 +41,13 @@ class CleanupJob extends AbstractQueuedJob implements QueuedJob {
 	);
 
 	/**
+	 * Database query limit
+	 * @config
+	 * @var integer
+	 */
+	private static $query_limit = 100000;
+
+	/**
 	 * Check whether is enabled or not for BC
 	 * @config
 	 * @var boolean
@@ -79,6 +86,13 @@ class CleanupJob extends AbstractQueuedJob implements QueuedJob {
 	 * Clear out stale jobs based on the cleanup values
 	 */
 	public function process() {
+		// construct limit statement if query_limit is valid int value
+		$limit = '';
+		$query_limit = $this->config()->get('query_limit');
+		if (is_numeric($query_limit) && $query_limit >= 0) {
+			$limit = ' LIMIT ' . ((int)$query_limit);
+		}
+
 		$statusList = implode('\', \'', $this->config()->cleanup_statuses);
 		switch($this->config()->cleanup_method) {
 			// If Age, we need to get jobs that are at least n days old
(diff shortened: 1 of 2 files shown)


code/jobs/CleanupJob.php
```diff
@@ -70,52 +80,56 @@ public function getTitle() {
...
 					'SELECT "ID" 
-					FROM "QueuedJobDescriptor" 
-					WHERE "JobStatus" 
-					IN (\'' . $statusList . '\')
-					AND "LastEdited" < \'' . $cutOff .'\''
+                    FROM "QueuedJobDescriptor" 
+                    WHERE "JobStatus" 
+                    IN (\'' . $statusList . '\') 
+                    AND "LastEdited" < \'' . $cutOff . '\' 
+                    LIMIT ' . $this->config()->query_limit
```
ScopeyNZ: You could probably extract this into a variable and assert it's a valid integer first.

```php
$limit = '';
$configuration = $this->config()->get('query_limit');
if (is_numeric($configuration)) {
	$limit = 'LIMIT ' . ((int) $configuration);
}
```

JoshuaCarter: @ScopeyNZ Agreed, and I'd briefly considered that but wound up following the existing patterns. I've overwritten the prievious commit with this change (and because my IDE had wrecked havok on the formatting).

code/jobs/CleanupJob.php
```diff
@@ -79,6 +86,13 @@ public function getJobType() {
 	 * Clear out stale jobs based on the cleanup values
 	 */
 	public function process() {
+		// construct limit statement if query_limit is valid int value
+		$limit = '';
+		$query_limit = $this->config()->get('query_limit');
+		if (is_numeric($query_limit) && $query_limit >= 0) {
```
nglasl: This PR looks okay to me, I assume you're hitting an issue where it's trying to clean up too many records at once? The only minor thing that jumps out to me is the `>= 0`. Is 0 a reasonable limit? I guess there may be a use case where someone wants to completely disable this job without actually stopping it from running, so setting it to 0 might make sense.

JoshuaCarter: Yea, I mostly just wanted to disallow negative integers and saw no particular reason to disallow 0. Thanks for review, I'll merge it in.

JoshuaCarter:
@ScopeyNZ I've made a release for this change: https://github.com/symbiote/silverstripe-queuedjobs/releases/tag/3.1.2

@nglasl tells me you may have preferences as to how this change is merged into other branches? How would you proceed from here, and I hope I haven't stepped on any toes.

JoshuaCarter:
@nglasl note that this is included in my 3.1.3 release (including description) so it may not be worth producing a separate 3.1.2 release for this anymore :)

JoshuaCarter:
@ScopeyNZ just a reminder about this, namely the 3.1.2 tag pointing to 4.x on master, and the question about this change going into other branches, etc.

robbieaverill:
Yep, sorry - looks like this was totally my bad. I think the best thing to do is to release a new 3.1.x tag on the correct release line, which will mean composer will choose it over 3.1.2.

We could delete the tag, but we may risk breaking people's sites if we do that

robbieaverill:
@JoshuaCarter in terms of merging up - I'll do this now for you

PR #226, opened 2018-11-28
FIX Remove _config.php check for folder name - it is a vendor module now

robbieaverill:


robbieaverill:
I think this snuck in from a SS3 merge up. This code doesn't exist in the stable SS4 releases, so this is fixing that merge issue.

PR #229, opened 2019-01-23
Fix(QueuedJobService): Valid $messages data being overwritten

JoshuaCarter:
Fixes issue https://github.com/symbiote/silverstripe-queuedjobs/issues/228 for 3.0 (and above after merging up).

### DIFF
--- a/code/services/QueuedJobService.php
+++ b/code/services/QueuedJobService.php
@@ -206,16 +206,9 @@ class QueuedJobService {
 		$jobData = @unserialize($jobDescriptor->SavedJobData);
 		$messages = @unserialize($jobDescriptor->SavedJobMessages);
 
-		if (!$jobData) {
-			// SS's convert:: function doesn't do this detection for us!!
-			if (function_exists('json_decode')) {
-				$jobData = json_decode($jobDescriptor->SavedJobData);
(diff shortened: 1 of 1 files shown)


code/services/QueuedJobService.php
```diff
@@ -206,15 +206,10 @@ protected function copyDescriptorToJob($jobDescriptor, $job) {
...
-			if (function_exists('json_decode')) {
-				$jobData = json_decode($jobDescriptor->SavedJobData);
-				$messages = json_decode($jobDescriptor->SavedJobMessages);
-			} else {
-				$jobData = Convert::json2obj($jobDescriptor->SavedJobData);
-				$messages = Convert::json2obj($jobDescriptor->SavedJobMessages);
-			}
+		// try decoding as json if null
+		if (function_exists('json_decode')) {
+			$jobData = $jobData ?: json_decode($jobData);
```
robbieaverill: Does this need to be decoding `$jobDescriptor->SavedJobData`?  Otherwise if `$jobData` is false from a failure in `unserialize()` then `json_decode(false)` won't help us

JoshuaCarter: *cough* I don't know what you mean >_>

PR #237, opened 2019-03-05
NEW: Improve job list in admin/queuedjobs

sminnee:
- Translate the job type from a number to a string
 - Add “finished”, which can be useful for seeing job duration

### DIFF
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -129,9 +129,10 @@ class QueuedJobDescriptor extends DataObject
             'JobTitle' => _t(__CLASS__ . '.TABLE_TITLE', 'Title'),
             'Created' => _t(__CLASS__ . '.TABLE_ADDE', 'Added'),
             'JobStarted' => _t(__CLASS__ . '.TABLE_STARTED', 'Started'),
+            'JobFinished' => _t(__CLASS__ . '.TABLE_FINISHED', 'Finished'),
 //          'JobRestarted' => _t(__CLASS__ . '.TABLE_RESUMED', 'Resumed'),
             'StartAfter' => _t(__CLASS__ . '.TABLE_START_AFTER', 'Start After'),
-            'JobType' => _t(__CLASS__ . '.JOB_TYPE', 'Job Type'),
+            'JobTypeString' => _t(__CLASS__ . '.JOB_TYPE', 'Job Type'),
             'JobStatus' => _t(__CLASS__ . '.TABLE_STATUS', 'Status'),
             'LastMessage' => _t(__CLASS__ . '.TABLE_MESSAGES', 'Message'),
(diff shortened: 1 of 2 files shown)


src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -284,6 +285,21 @@ public function getTitle()
...
+     * Return a string representation of the numeric JobType
+     * @return string
+     */
+    public function getJobTypeString()
+    {
+        $map = [
+            QueuedJob::IMMEDIATE => 'Immediate',
+            QueuedJob::QUEUED => 'Queued',
+            QueuedJob::LARGE => 'Large',
+        ];
```
robbieaverill: Do you think we should make these translatable as well?

sminnee: Yeah, makes sense.

sminnee:
The "finished" column is not hugely useful unless max_finished_jobs_age is reasonably long. By default it's 300 (5 minutes). On the project I have most to do with queuedjobs on, I set this to 7200 (2 hours) which gives you a better view of recent jobs in the admin.

Should we increase the default value of max_finished_jobs_age too?

robbieaverill:
I think it'd totally depend on what kind of project you have, which would probably be how many jobs you run regularly. I think 5 mins is probably useless for most people so yeah I'd be happy to increase it to a couple of hours.

src/DataObjects/QueuedJobDescriptor.php
sminnee: Done.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -284,6 +285,30 @@ public function getTitle()
...
+    /**
+     * Return a map of numeric JobType values to localisable string representations.
+     * @return array
+     */
+    public function getJobTypeValues()
+    {
+        return [
+            QueuedJob::IMMEDIATE => _t(QueuedJob::class . '.TYPE_IMMEDIATE', 'Immediate'),
+            QueuedJob::QUEUED => _t(QueuedJob::class . '.TYPE_QUEUED', 'Queued'),
+            QueuedJob::LARGE => _t(QueuedJob::class . '.TYPE_LARGE', 'Large'),
```
robbieaverill: Unfortunately `Some::class` doesn't work in the text collector. Will need to be `_t('Symbiote\\QueuedJobs\\Services\\QueuedJob',`

sminnee: Doh, you're right https://github.com/silverstripe/silverstripe-framework/issues/7647

I might just use `__class__`; these values will be attached to QueuedJobDescriptor but that's not a showstopper?

robbieaverill: Yep sounds good

sminnee: Done; squashed in.

PR #239, opened 2019-03-27
FIX: Add check for SubsiteID

Leapfrognz:


Leapfrognz:
This will probably need to go into a SS3 release as well.

src/Services/QueuedJobService.php
```diff
@@ -2,27 +2,27 @@
...
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Extensible;
+use SilverStripe\Core\Convert;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Control\Email\Email;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\Controller;
+use Psr\Log\LoggerInterface;
+use Exception;
```
ScopeyNZ: It looks like your IDE is set to order imports in reverse (`Z-a`). Super minor request but do you mind reverting this part of your commit?

src/Services/QueuedJobService.php
```diff
@@ -664,7 +664,7 @@ public function runJob($jobId)
                 // have we stalled at all?
                 $stallCount = 0;
 
-                if ($job->SubsiteID && class_exists(Subsite::class)) {
+                if (class_exists(Subsite::class) && isset($job->SubsiteID)) {
```
ScopeyNZ: Maybe `!empty($job->SubsiteID)`? I'm not sure if `$job->SubsiteID === 0` was a valid path before but now that would return `true` instead of `false` previously.

src/Services/QueuedJobService.php
Leapfrognz: Decsendist! 😛

Leapfrognz:
Updates - removed resort of use statements (sorry) and changed to !empty($job->SubsiteID)

ScopeyNZ:
One last comment - do you want to target a patch release for this bugfix? Currently `master` is the next minor release of the `4` line.

If you want to target a patch release then you might want to target `4.3`? I can explain a little more about our branching strategy if you're unsure?

Leapfrognz:
> One last comment - do you want to target a patch release for this bugfix? Currently `master` is the next minor release of the `4` line.
> 
> If you want to target a patch release then you might want to target `4.3`? I can explain a little more about our branching strategy if you're unsure?

Sure, if you think that's best. But we need it for a SS3 project for the time being, not sure on that process?

ScopeyNZ:
Well you can target the next patch release of whatever version of this module you're currently using... If you're using `3.1.3` then you can target `3.1`. You'll also have to rebase to ensure you don't pick up any commits from the `master` branch. `git rebase origin/4.3; git push -f`.

PR #240, opened 2019-04-03
Add real-time logs when using `LoggerInterface`

bergice:


src/Services/QueuedJobLogger.php
```diff
@@ -0,0 +1,41 @@
...
+use Psr\Log\AbstractLogger;
+use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
+
+/**
+ * Writes log output to a job descriptor
+ */
+class QueuedJobLogger extends AbstractLogger
+{
+    /** @var QueuedJob */
+    public $job;
```
dnsl48: Do we actually need these class properties to be public?

It's usually a thumbs-up rule to give as strict access as possible by default and only open things when it's really necessary. Otherwise, we introduce it to the public APIs which we'll have to support for the whole lifecycle of the module and won't be able to break it until the next major version release.

robbieaverill: Yep agreed. These can both be protected with a getter for external access

src/Services/QueuedJobLogger.php
```diff
@@ -0,0 +1,41 @@
...
+     * @param string $message
+     * @param array $context
+     *
+     * @return void
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    public function log($level, $message, array $context = array())
+    {
+        $this->job->addMessage($message);
+        $this->jobDescriptor->SavedJobMessages = serialize($this->job->getJobData()->messages);
```
robbieaverill: Can you please use JSON encoding instead?

src/Services/QueuedJobLogger.php
```diff
@@ -0,0 +1,41 @@
...
+     * Logs with an arbitrary level.
+     *
+     * @param mixed $level
+     * @param string $message
+     * @param array $context
+     *
+     * @return void
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    public function log($level, $message, array $context = array())
```
robbieaverill:
```suggestion
    public function log($level, $message, array $context = [])
```

(stylistic suggestion)

src/Services/QueuedJobService.php
```diff
@@ -708,7 +708,11 @@ public function runJob($jobId)
                     }
 
                     if (!$broken) {
-                        // Collect output as job messages as well as sending it to the screen
+                        // Inject real-time logger
+                        $queuedJobLogger = new QueuedJobLogger($job, $jobDescriptor);
```
robbieaverill: Please use Injector here to allow it to be extensible:

```suggestion
                        $queuedJobLogger = QueuedJobLogger::create($job, $jobDescriptor);
```

(Requires suggestion above)

src/Services/QueuedJobLogger.php
```diff
@@ -0,0 +1,41 @@
...
+namespace Symbiote\QueuedJobs\Services;
+
+use Psr\Log\AbstractLogger;
+use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
+
+/**
+ * Writes log output to a job descriptor
+ */
+class QueuedJobLogger extends AbstractLogger
+{
```
robbieaverill:
```suggestion
{
    use Injectable;
```

To allow Injector to instantiate

src/Services/QueuedJobService.php
```diff
@@ -708,7 +708,11 @@ public function runJob($jobId)
...
+                        Injector::inst()->registerService($queuedJobLogger, LoggerInterface::class);
```
robbieaverill: I'm not convinced that this is a good decision - you're replacing the LoggerInterface that exists already with your own implementation. What you should do is push new handlers into the stack of the existing LoggerInterface.

robbieaverill:
Also as an enhancement can you please target the PR at the next minor branch?

src/Services/QueuedJobService.php
dnsl48: Yeah, I thought about this as well. Don't like replacement of the global logger, but I could not come up with how to solve this.
I guess since Monolog is a framework requirement we could switch here to `Monolog\Logger` and simply add another handler then?

robbieaverill: Is the objective to use the core error logger? If so, just access it via its service name. Note the PR I mentioned earlier has been merged now, so the logger service name will be `LoggerInterface.core` or something from 4.4 onwards, which will make my team's life a little more complicated for things like this, but anyway...

You could also define your own logger - see https://docs.silverstripe.org/en/4/developer_guides/debugging/error_handling/

I'd suggest using the core error logger for now. `Injector::inst()->get(LoggerInterface::class)->info('hi')` etc

dnsl48: Yeah, agree, it makes more sense for this service to use the core logger rather than registering a new one.
@bergice I guess that's been done in the first place just because 4.3 does not provide a default logger?

bergice:
> Also as an enhancement can you please target the PR at the next minor branch?

@robbieaverill  There's no `4` branch at the moment, how do I create it?

robbieaverill:
Ah it's master in this repo

bergice:
This should use a log handler now.

src/Services/QueuedJobHandler.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __construct(QueuedJob $job, QueuedJobDescriptor $jobDescriptor)
+    {
+        $this->job = $job;
+        $this->jobDescriptor = $jobDescriptor;
+    }
+
+    /**
+     * @return QueuedJob
+     */
+    public function getJob(): QueuedJob
```
dnsl48: yeah, that would be cool, but we still have to support php 5.6 for some while :)

dnsl48:
@robbieaverill all good now?

robbieaverill:
The builds are red... I'll review the code shortly

dnsl48:
Yeah, that's styling
```
FILE: ...s/build/symbiote/silverstripe-queuedjobs/src/Services/QueuedJobService.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 741 | WARNING | Line exceeds 120 characters; contains 121 characters
--------------------------------------------------------------------------------
```
cc @bergice

bergice:
> Yeah, that's styling
> 
> ```
> FILE: ...s/build/symbiote/silverstripe-queuedjobs/src/Services/QueuedJobService.php
> --------------------------------------------------------------------------------
> FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
> --------------------------------------------------------------------------------
>  741 | WARNING | Line exceeds 120 characters; contains 121 characters
> --------------------------------------------------------------------------------
> ```
> cc @bergice

Not sure what is going on here, but that line is not over 120 characters: https://github.com/open-sausages/silverstripe-queuedjobs/blob/pulls/4/logging/src/Services/QueuedJobService.php#L741

robbieaverill:
Might be a phpcs glitch. They release some new bugfixes yesterday which might be related. I've restarted the builds.

sminnee:
Is it intentional that this go into `master`?

robbieaverill:
@sminnee yeah if it's an enhancement

robbieaverill:
I tested the phpcs error manually by recreating the Travis build steps, and it doesn't occur for me. I'll re-run the builds again.

robbieaverill:
I've rebased and squashed the commits at the same time

src/Services/QueuedJobLogger.php
robbieaverill: This isn't resolved

src/Services/QueuedJobService.php
```diff
@@ -731,7 +732,17 @@ public function runJob($jobId)
                     }
 
                     if (!$broken) {
-                        // Collect output as job messages as well as sending it to the screen
+                        // Inject real-time log handler
+                        $logger = Injector::inst()->get(LoggerInterface::class);
+                        if ($logger instanceof Logger) {
+                            $logger->pushHandler(new QueuedJobHandler($job, $jobDescriptor));
```
```suggestion
                            $logger->pushHandler(QueuedJobHandler::create($job, $jobDescriptor));
```

robbieaverill: Do you think it might be worth checking whether the Logger already has this handler? I'm thinking about if you have a service running multiple jobs in the same process, it may end up stacking QueuedJobHandler instances into the logger, and aggregating old job information that isn't relevant after it's completed, e.g.:

- Run job 1

- Run job 1 (irrelevant now)
- Run job 2

- Run job 1 (irrelevant now)
- Run job 2 (irrelevant now)
- Run job 3

robbieaverill:
OK I can reproduce the phpcs failure now. Looks like it's run against the merged version of the code from the PR rather than what's in the fork you're PR'ing from. I'll patch it now.

src/Services/QueuedJobService.php
bergice: Good point, I'll add this now.

src/Services/QueuedJobLogger.php
bergice: Sorry, should have noted that we can't do this. `SavedJobMessages` already uses `serialize`/`unserialise`, which means that there is also data encoded that way.

robbieaverill:
LGTM. I'll squash on merge to remove all the (changelog) irrelevant commit messages and add a NEW prefix

src/Services/QueuedJobLogger.php
robbieaverill: It actually tries to do both, we are trying to prefer JSON serialisation in general nowadays. It's OK for now though.

src/Services/QueuedJobHandler.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param  array $record
+     * @return void
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    protected function write(array $record)
+    {
+        $this->job->addMessage($record['message']);
+        $this->jobDescriptor->SavedJobMessages = serialize($this->job->getJobData()->messages);
+        $this->jobDescriptor->write();
```
chillu: @bergice That's one database write per log entry. Which can be just as expensive as the task it's actually running. In the case which triggered this, that would be moving potentially tens of thousands of files. I think we just made this task sustantially slower by introducing logging. Your local database won't be a good test case for this, remove databases have at least 1ms roundtrip. Have you looked at Monolog-level batching?

bergice: Good point, I've implemented this using `BufferHandler`.

PR #241, opened 2019-04-17
Batch log output

bergice:
Batches output using BufferHandler so we don't constantly write logs to the database.


robbieaverill:
Didn't we just merge a bunch of this with #240? Perhaps it needs a rebase. Would be nice to see the updated commits only if possible, could you also please give your PR a relevant title and description?

bergice:
Moving to https://github.com/symbiote/silverstripe-queuedjobs/pull/242

PR #243, opened 2019-05-08
Fix mutex

stephenmcm:


stephenmcm:
Cheers @robbieaverill. We should also bring this fix through into master, while technically the default runner validates the job, this isn't the case in [silverstripe-sqs-jobqueue](https://github.com/nyeholt/silverstripe-sqs-jobqueue) when running queued instead of immediate jobs. While somewhat of an edge case the grabMutex method should probably be as robust as possible.

robbieaverill:
Sure thing. We don't usually merge up between major versions, so I'll cherry-pick this commit instead.

PR #244, opened 2019-05-09
minor - log information to info instead of error

torleif:
When queuing default job, log to info() instead of error()

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -422,7 +422,7 @@ class QueuedJobService
                     $jobConfig['filter']
                 ));
                 if (!$job->count()) {
-                    $this->getLogger()->error(
+                    $this->getLogger()->info(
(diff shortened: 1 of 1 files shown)


robbieaverill:
This won't be a problem in SilverStripe 4.4, since LoggerInterface becomes a user logger by default instead of a core logger. I guess until that becomes a minimum dependency in this module we should do this though.

torleif:
@robbieaverill we've noticed that the CWP logger sends emails on errors. Is there any chance if this getting merged in?

robbieaverill:
cc @chrispenny @mfendeksilverstripe I remember you were doing something around this in your project

mfendeksilverstripe:
@robbieaverill That's correct, we had to rewrite a lot of the logging related code because of the massive debug information which was needed to cover our customization of the module. We have a plan to clean up our project code and feedback some functionality back to the module ;-).

PR #246, opened 2019-05-22
3.0 - Check Subsite class exists before checking SubsiteID

emteknetnz:
Ensure that that Subsite class exists before checking for SubsiteID

### DIFF
--- a/code/services/QueuedJobService.php
+++ b/code/services/QueuedJobService.php
@@ -559,7 +559,7 @@ class QueuedJobService {
 				// have we stalled at all?
 				$stallCount = 0;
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
I've restarted the most recent 3.0 branch build to ensure that the test failure here is pre-existing

robbieaverill (approved):

robbieaverill:
Latest 3.0 build is green, this change is breaking the tests

robbieaverill (changes requested):
Needs the tests fixed

robbieaverill (approved):
Re-ran tests again, I think this is the "tests ran on the hour so failed" timing issue. We've fixed that in SS4.

PR #247, opened 2019-05-23
Extensibility options improvements.

mfendeksilverstripe:
Multiple extensibility options added:

* `getRunAsMemberID` configuration - this allows the jobs to be run without a user (job level configurable)
* `max_init_jobs` option added - prevents too many jobs to be in the `init` state
* `updateJobDescriptorBeforeQueued ` extension point added - this allows custom DB columns to be populated before the job is queued (used for data that can't be present in the serialized job data like identifiers for example)
* `updateJobDescriptorAndJobOnCompletion` extension point added - this is useful for `dependent` jobs, i.e. jobs that are queued only after other jobs finish
* PHP 7 exception handling added - type errors were not handled properly before

Compatibility notes:

* jobs that do not extend `AbstractQueuedJob` need to add the `getRunAsMemberID` function
* if userID was `0` during `queueJob` it won't be replaced by current user ID, use `null` instead if you want previous behaviour

### DIFF
--- a/src/Services/AbstractQueuedJob.php
+++ b/src/Services/AbstractQueuedJob.php
@@ -117,6 +117,11 @@ abstract class AbstractQueuedJob implements QueuedJob
         return QueuedJob::QUEUED;
     }
 
+    public function getRunAsMemberID()
+    {
+        return null;
+    }
+
     /**
      * Performs setup tasks the first time this job is run.
      *
--- a/src/Services/QueuedJob.php
+++ b/src/Services/QueuedJob.php
@@ -85,6 +85,23 @@ interface QueuedJob
     public function getJobType();
 
     /**
+     * Specifies what user ID should be when running the job
+     * valid values:
+     * null - (default) - run the job as current user
+     * 0 - run the job without a user
+     * greater than zero - run the job as a specific user
+     *
+     * This is useful in situations like:
+     * - a job needs to always run without a user (like a static cache job)
+     * - a job needs to run as a specific user (for example data migration job)
+     *
+     * Note that this value can be overriden in the @see QueuedJobService::queueJob()
+     *
+     * @return int|null
+     */
+    public function getRunAsMemberID();
+
+    /**
      * A job is run within an external processing loop that will call this method while there are still steps left
      * to complete in the job.
      *
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -18,6 +18,7 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBDatetime;
+use SilverStripe\ORM\ValidationException;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Security;
 use SilverStripe\Subsites\Model\Subsite;
@@ -95,6 +96,21 @@ class QueuedJobService
     private static $disable_health_check = false;
 
     /**
+     * Maximum number of jobs that can be initialised at any one time.
+     *
+     * Prevents too many jobs getting into this state in case something goes wrong with the child processes.
+     * We shouldn't have too many jobs in the initialising state anyway.
+     *
+     * Valid values:
+     * 0 - unlimited (default)
+     * greater than 0 - maximum number of jobs in initialised state
+     *
+     * @var int
+     * @config
+     */
+    private static $max_init_jobs = 0;
+
+    /**
      * Timestamp (in seconds) when the queue was started
      *
      * @var int
@@ -159,16 +175,12 @@ class QueuedJobService
     /**
      * Adds a job to the queue to be started
      *
-     * Relevant data about the job will be persisted using a QueuedJobDescriptor
-     *
-     * @param QueuedJob $job
-     *          The job to start.
-     * @param $startAfter
-     *          The date (in Y-m-d H:i:s format) to start execution after
-     * @param int $userId
-     *          The ID of a user to execute the job as. Defaults to the current user
-     *
+     * @param QueuedJob $job The job to start.
+     * @param null|string $startAfter The date (in Y-m-d H:i:s format) to start execution after
+     * @param null|int $userId The ID of a user to execute the job as. Defaults to the current user
+     * @param null|int $queueName
      * @return int
+     * @throws ValidationException
      */
     public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $queueName = null)
     {
@@ -198,13 +210,31 @@ class QueuedJobService
         $jobDescriptor->Implementation = get_class($job);
         $jobDescriptor->StartAfter = $startAfter;
 
-        $runAsID = 0;
-        if ($userId) {
-            $runAsID = $userId;
-        } elseif (Security::getCurrentUser() && Security::getCurrentUser()->exists()) {
-            $runAsID = Security::getCurrentUser()->ID;
+        // no user provided - fallback to job user default
+        if ($userId === null) {
+            $userId = $job->getRunAsMemberID();
         }
+
+        // still no user - fallback to current user
+        if ($userId === null) {
+            if (Security::getCurrentUser() && Security::getCurrentUser()->exists()) {
+                // current user available
+                $runAsID = Security::getCurrentUser()->ID;
+            } else {
+                // current user unavailable
+                $runAsID = 0;
+            }
+        } else {
+            $runAsID = $userId;
+        }
+
         $jobDescriptor->RunAsID = $runAsID;
+
+        // use this to populate custom data columns before job is queued
+        // note: you can pass arbitrary data to your job and then move it to job descriptor
+        // this is useful if you need some data that needs to be exposed as a separate
+        // DB column as opposed to serialised data
+        $this->extend('updateJobDescriptorBeforeQueued', $jobDescriptor, $job);
 
         // copy data
         $this->copyJobToDescriptor($job, $jobDescriptor);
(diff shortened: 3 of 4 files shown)


src/Services/AbstractQueuedJob.php
```diff
@@ -117,6 +117,14 @@ public function getJobType()
         return QueuedJob::QUEUED;
     }
 
+    /**
+     * @return int|null
+     */
+    public function getRunAsMember()
```
robbieaverill: I was about to ask for a better docblock until I saw it in the QueuedJob interface. If you omit the docblock entirely here, PHPDoc will inherit it automatically. The way it is written now will override the interface's doc block

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
      */
     public function getJobType();
 
+    /**
+     * Specifies what user ID should be hwen runnning the job
```
robbieaverill:
```suggestion
     * Specifies what user ID should be when running the job
```

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
...
+     * valid values:
+     * null - (default) - run the job as current user
+     * 0 - run the job without a user
+     * greater than zero - run the job as a specifi user
```
robbieaverill:
```suggestion
     * greater than zero - run the job as a specific user
```

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
...
+     *
+     * thisis useful in situation like:
```
robbieaverill:
```suggestion
     * This is useful in situations like:
```

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
...
+     * - a job needs to always run without user (like static cache job)
```
robbieaverill:
```suggestion
     * - a job needs to always run without a user (like a static cache job)
```

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
...
+     * - a job needs to run as a specific user (for example data migration job)
+     *
+     * Note that this value can be overriden in the @see QueuedJobService::queueJob()
+     *
+     * @return int|null
+     */
+    public function getRunAsMember();
```
robbieaverill: Thoughts on calling this `getRunAsMemberID()` since it returns an ID rather than a Member instance?

src/Services/QueuedJobService.php
```diff
@@ -94,6 +95,20 @@ class QueuedJobService
      */
     private static $disable_health_check = false;
 
+    /**
+     * Maximum number of jobs that can be initialised at any one time
```
```suggestion
     * Maximum number of jobs that can be initialised at any one time.
     *
```

src/Services/QueuedJobService.php
```diff
@@ -94,6 +95,20 @@ class QueuedJobService
...
+     * prevents too many jobs getting into this state in case something goes wrong with the child processes
```
```suggestion
     * Prevents too many jobs getting into this state in case something goes wrong with the child processes.
```

src/Services/QueuedJobService.php
```diff
@@ -94,6 +95,20 @@ class QueuedJobService
...
+     * we shouldn't have too many jobs in the initialising state anyway
```
```suggestion
     * We shouldn't have too many jobs in the initialising state anyway.
```

src/Services/QueuedJobService.php
```diff
@@ -94,6 +95,20 @@ class QueuedJobService
...
+     *
+     * valid values:
```
```suggestion
     * Valid values:
```

src/Services/QueuedJobService.php
```diff
@@ -159,18 +174,15 @@ public function __construct()
...
+     * @param QueuedJob $job The job to start.
+     * @param null|string $startAfter The date (in Y-m-d H:i:s format) to start execution after
+     * @param null|int $userId The ID of a user to execute the job as. Defaults to the current user
+     * @param null|int $queueName
+     * @param array $options options which are passed to the updateJobDescriptorBeforeQueued extension point
      * @return int
+     * @throws ValidationException
      */
-    public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $queueName = null)
+    public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $queueName = null, array $options = [])
```
robbieaverill: Adding this would be a breaking change, which would mean it'll take longer to get released. Is there a way you could implement this without changing the method signature?

src/Services/QueuedJobService.php
```diff
@@ -232,6 +259,30 @@ public function startJob($jobDescriptor, $startAfter = null)
         }
     }
 
+    /**
+     * Check if maximum number of jobs are currently initialised.
+     *
+     * @codeCoverageIgnore
```
```suggestion
```

We don't make exclusions for code coverage in our open source modules, we prefer to be honest about the missing bits 😆

src/Services/QueuedJobService.php
```diff
@@ -232,6 +259,30 @@ public function startJob($jobDescriptor, $startAfter = null)
...
+     * @return bool
+     */
+    public function isAtMaxJobs()
+    {
+        $initJobsMax = static::config()->get('max_init_jobs');
```
```suggestion
        $initJobsMax = $this->config()->get('max_init_jobs');
```

src/Services/QueuedJobService.php
```diff
@@ -892,18 +943,16 @@ public function runJob($jobId)
...
-                    ]
-                );
-                $jobDescriptor->JobStatus =  QueuedJob::STATUS_BROKEN;
-                $this->extend('updateJobDescriptorAndJobOnException', $jobDescriptor, $job, $e);
-                $jobDescriptor->write();
+                // PHP 5.6 exception handling
+                $this->handleBrokenJobException($jobDescriptor, $job, $e);
+                $broken = true;
+            } catch (\Throwable $e) {
+                // PHP 7 exception handling
```
robbieaverill: Technically the majority of PHP 7 exceptions would be caught in the previous catch, right? Would it be more honest to say "PHP 7 Error handling" ?

src/Services/QueuedJobService.php
```diff
@@ -917,6 +966,26 @@ public function runJob($jobId)
         return !$broken;
     }
 
+    /**
+     * @param QueuedJobDescriptor $jobDescriptor
+     * @param QueuedJob $job
+     * @param $e
```
```suggestion
     * @param Exception|\Throwable $e
```

src/Services/QueuedJobService.php
```diff
@@ -917,6 +966,26 @@ public function runJob($jobId)
...
+     * @throws ValidationException
```
robbieaverill: We don't usually document exceptions unless the function explicitly throws them itself, I guess this comes from the write() call, but maybe omit it from here

src/Tasks/Engines/DoormanRunner.php
```diff
@@ -47,6 +48,13 @@ public function getDefaultRules()
      */
     public function runQueue($queue)
     {
+        // check if queue can be processed
+        $service = QueuedJobService::singleton();
+        if ($service->isAtMaxJobs()) {
+            $service->getLogger()->error('Not processing queue as jobs are at max initialisation limit.');
```
robbieaverill: Note that #244 reduced this to an `info` level. I have just merged those branches up so, so you'll probably get conflicts in the other method refactor you did above, but maybe you could use `->info()` here instead?

mfendeksilverstripe:
@robbieaverill Thanks for the feedback, I pushed some new changes, please review ;-).

src/Services/QueuedJob.php
```diff
@@ -84,6 +84,23 @@ public function prepareForRestart();
...
+     *
+     * This is useful in situations like:
+     * - a job needs to always run without a user (like a static cache job)
+     * - a job needs to run as a specific user (for example data migration job)
+     *
+     * Note that this value can be overriden in the @see QueuedJobService::queueJob()
+     *
+     * @return int|null
+     */
+    public function getRunAsMemberID();
```
robbieaverill: Sorry, I didn't pick this up as a breaking change. While AbstractQueuedJob implements this interface, not all jobs do (including those in this module). We'll need to remove this from the interface and adjust the call logic to do a `method_exists()` check

mfendeksilverstripe: @robbieaverill I'm not sure I follow. As far as I can see, all jobs in this module extend the `AbstractQueuedJob` which has the function defined and uses the interface. COuld you please clarify what you mean by breaking change exactly?

robbieaverill: Yeah sorry, it's not in this module but in fulltextsearch - see `SearchUpdateCommitJobProcessor` as an example that implements the interface but doesn't extend the abstract base class

ScopeyNZ: Unfortunately @mfendeksilverstripe adding methods to interfaces is a breaking change :(.

ScopeyNZ: You may introduce new interfaces though

mfendeksilverstripe: I see I haven't checked other module. I like the idea of a new Interface. Should I create a PR for that, or are you going to hotfix it?

ScopeyNZ: We won't release until it's resolved. You're welcome to create a PR for it otherwise I'll try to get around to it at some point today.

Yeah we can't even introduce a method to an interface even if we go and update all the open source code to implement the new method. We can't see all the closed source code that might implement this interface too 😉

mfendeksilverstripe: I'll have a look at it today. It shouldn't be too much work.

mfendeksilverstripe: Open PR with the fix: https://github.com/symbiote/silverstripe-queuedjobs/pull/262

PR #248, opened 2019-05-23
Queued only tasks configuration added.

mfendeksilverstripe:
* Added an option to have queued only tasks via configuration
* this is useful for tasks that take too long to run via browser
* especially useful when task produce output as this will get captured in the queued job instead of timing out via browser


robbieaverill:
LGTM, let's target a new minor release though since it's a feature. Can you change the target to master and rebase please? You can do this through the GitHub UI (then rebase locally and force push)

mfendeksilverstripe:
@robbieaverill Base changed to `master ;-)

PR #249, opened 2019-05-30
Allow queued_job_admin_email to be disabled

wilr:
At the moment setting this to false does not work since it defaults to the admin address. This adds support to disable it by explicitly setting it to false.


ScopeyNZ:
LGTM. Needs a rebase now though.

wilr:
@ScopeyNZ rebased.

PR #250, opened 2019-06-05
Log severity and original date in queue messages

chillu:
The severity was set to INFO, which is a bug: It masks the actual severity of the log message,
for example ERROR or ALERT.

~Recording the datetime of addMessage() as the current datetime is also a bug:
It masks the actual date when the message was recorded, which could be many seconds
earlier due to message batching. This can in turn mask issues around slow executions,
e.g. during the file migration tasks for which this layer was introduced.
Note that the QueuedJob interface which AbstractQueuedJob implements does not have this new third argument. There's already precedent for adding optional arguments to the interface with $severity.~

### DIFF
--- a/src/Services/QueuedJobHandler.php
+++ b/src/Services/QueuedJobHandler.php
@@ -57,7 +57,7 @@ class QueuedJobHandler extends AbstractProcessingHandler
     public function handleBatch(array $records)
     {
         foreach ($records as $record) {
-            $this->job->addMessage($record['message']);
+            $this->job->addMessage($record['message'], $record['level_name'], $record['datetime']);
         };
         $this->jobDescriptor->SavedJobMessages = serialize($this->job->getJobData()->messages);
 


robbieaverill:
Just a note to be careful with log severity levels. Before SilverStripe 4.4 anything higher than info level would be caught in the core error handlers, which is why we've been forced to use low levels here.

chillu:
lol my proof was the wrong way around, it *does* throw ...

```php
<?php
class A {
	function foo($bar, $baz = null) {}
};
class B extends A {
	function foo($bar) {}
};
$b = new B();
$b->foo("bar");
```

```
PHP Warning:  Declaration of B::foo($bar) should be compatible with A::foo($bar, $baz = NULL) in test.php on line 7
```

So we're relying on the fact that `AbstractQueuedJob->addMessage()` hasn't been overwritten in subclasses. It's not the case in the module itself, and I can't see any uses when [searching across Github](https://github.com/search?q=AbstractQueuedJob+%22function+addMessage%22&type=Code). But it's still a bit dodgy. Just working on an alternative approach.

chillu:
I was thinking about adding `AbstractQueuedJob->addMessageRaw(string $message)`, which doesn't attempt to do any formatting (e.g. defer it to `QueuedJobHandler->getFormatter()->format()`).

As a reminder, here's what `addMessage()` does:

```php
public function addMessage($message, $severity = 'INFO', \DateTime $datetime = null)
{
    $dbDatetime = $datetime ? (new DBDatetime())->setValue($datetime->getTimestamp()) : DBDatetime::now();
    $severity = strtoupper($severity);
    $this->messages[] = '[' . $dbDatetime->Rfc2822() . "][$severity] $message";
}
```

Ideally we'd just shift all logging in that implementation to a `Logger` implementation, rather than patching one in via `QueuedJobService`, but that's a pretty big change to the module. OK, I'm just reverting the date change, it's too low value to get into a whole refactoring discussion about it.

chillu:
> Just a note to be careful with log severity levels. Before SilverStripe 4.4 anything higher than info level would be caught in the core error handlers, which is why we've been forced to use low levels here.

Interesting, thanks for raising that. I don't think it influences this PR behaviour (it's just strings and log lines). We're started using `$this->logger->error()` etc in `MigrateFileTask` in 4.4, so that's not an issue since all that logic requires 4.4 anyway.

PR #255, opened 2019-06-20
Queued jobs pause setting added.

mfendeksilverstripe:
**Queued jobs pause setting added**

* this new feature adds a UI to `SiteConfig` which allows the jobs queue to be paused
* this is a `soft` pause, which means it actually lets the running jobs to finish, but prevents new jobs from starting
* very useful for planned downtime
* note that the maintenance lock status is not saved in DB which is intentional
* this is because downtime can be caused by DB restore operation which changes the whole DB
* pausing the jobs the DB restore is very helpful when moving `prod` DB to other `non-prod` environments as it prevents possible data corruption (the jobs can start running before the DB restoration is complete)
* also, some jobs may need to be cleared or updated when moved to another environment (such as absolute URLs), this can be done before the jobs are unpaused
* this feature can be enabled / disabled and is disabled by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,22 @@ Symbiote\QueuedJobs\Jobs\CleanupJob:
     - Complete
 ```
 
+## Jobs queue pause setting
+
+It's possible to enable a setting which allows the pausing of the queued jobs processing. To enable it, add following code to your config YAML file:
+
+```yaml
+Symbiote\QueuedJobs\Services\QueuedJobService:
+  lock_file_enabled: true
+  lock_file_path: '/shared-folder-path'
+```
+
+`Queue settings` tab will appear in the CMS settings and there will be an option to pause the queued jobs processing. If enabled, no new jobs will start running however, the jobs already running will be left to finish.
+ This is really useful in case of planned downtime like queue jobs related third party service maintenance or DB restore / backup operation.
+
+Note that this maintenance lock state is stored in a file. This is intentionally not using DB as a storage as it may not be available during some maintenance operations.
+Please make sure that the `lock_file_path` is pointing to a folder on a shared drive in case you are running a server with multiple instances.
+
 ## Health Checking
 
 Jobs track their execution in steps - as the job runs it increments the "steps" that have been run. Periodically jobs
--- a/src/Extensions/MaintenanceLockExtension.php
+++ b/src/Extensions/MaintenanceLockExtension.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Symbiote\QueuedJobs\Extensions;
+
+use SilverStripe\Forms\CheckboxField;
+use SilverStripe\Forms\FieldList;
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\SiteConfig\SiteConfig;
+use Symbiote\QueuedJobs\Services\QueuedJobService;
+
+/**
+ * Class MaintenanceLockExtension
+ * Adds a maintenance lock UI to SiteConfig
+ *
+ * @property SiteConfig|$this owner
+ * @package Symbiote\QueuedJobs\Extensions
+ */
+class MaintenanceLockExtension extends DataExtension
+{
+    /**
+     * @param FieldList $fields
+     */
+    public function updateCMSFields(FieldList $fields)
+    {
+        if (!QueuedJobService::config()->get('lock_file_enabled')) {
+            return;
+        }
+
+        $fields->addFieldsToTab('Root.QueueSettings', [
+            $lockField = CheckboxField::create(
+                'MaintenanceLockEnabled',
+                _t(__CLASS__ . '.LOCK_ENABLED', 'Maintenance Lock Enabled'),
+                QueuedJobService::singleton()->isMaintenanceLockActive()
+            ),
+        ]);
+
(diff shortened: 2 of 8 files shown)


mfendeksilverstripe:
@robbieaverill Fixed the unit tests. Is any other feedback required?

README.md
```diff
@@ -244,6 +244,22 @@ Symbiote\QueuedJobs\Jobs\CleanupJob:
...
+It's possible to enable a setting which allows the pausing of the queued jobs processing. To enable it, add following code to your config YAML file:
+
+```yaml
+Symbiote\QueuedJobs\Services\QueuedJobService:
+  lock_file_enabled: true
+  lock_file_path: '/shared-folder-path'
+```
+
+`Queue settings` tab will appear in the CMS settings and there will be an option to pause the queued jobs processing. If enabled, no new jobs will start running however, the jobs already running will be left to finish.
+ This is really useful in case of planned downtime like queue jobs related third party service maintenance or DB restore / backup operation.
```
robbieaverill: @mateusz @stojg this might be useful for platform integrations during deployments

src/Extensions/MaintenanceLockExtension.php
```diff
@@ -0,0 +1,57 @@
...
+    public function updateCMSFields(FieldList $fields)
+    {
+        if (!QueuedJobService::config()->get('lock_file_enabled')) {
+            return;
+        }
+
+        $fields->addFieldsToTab('Root.QueueSettings', [
+            $lockField = CheckboxField::create(
+                'MaintenanceLockEnabled',
+                'Maintenance Lock Enabled',
```
robbieaverill: We should make this translatable:

```suggestion
                _t(__CLASS__ . '.LOCK_ENABLED', 'Maintenance Lock Enabled'),
```

src/Extensions/MaintenanceLockExtension.php
```diff
@@ -0,0 +1,57 @@
...
+                QueuedJobService::isMaintenanceLockActive()
```
robbieaverill: This might be personal preference, but I think this should be a non-static method on the `QueuedJobService` singleton

src/Extensions/MaintenanceLockExtension.php
```diff
@@ -0,0 +1,57 @@
...
+            ),
+        ]);
+
+        $lockField->setDescription('Enable maintenance lock to prevent new queued jobs from being started');
```
robbieaverill: Can you make this translatable too please?

src/Extensions/MaintenanceLockExtension.php
```diff
@@ -0,0 +1,57 @@
...
+        if (!QueuedJobService::config()->get('lock_file_enabled')) {
+            return;
+        }
+
+        if ($value && !QueuedJobService::isMaintenanceLockActive()) {
+            QueuedJobService::enableMaintenanceLock();
+        }
+
+        if (!$value && QueuedJobService::isMaintenanceLockActive()) {
+            QueuedJobService::disableMaintenanceLock();
```
robbieaverill: As with the previous comment - can you make these non-static state setters on a singleton instead?

src/Tasks/Engines/DoormanRunner.php
```diff
@@ -78,6 +78,10 @@ public function runQueue($queue)
         $descriptor = $this->getNextJobDescriptorWithoutMutex($queue);
 
         while ($manager->tick() || $descriptor) {
+            if (QueuedJobService::isMaintenanceLockActive()) {
+                return;
```
robbieaverill: Do you want to add some logging here, if there's a logger available? Just something like `->info('Skipped queued job descriptor since maintenance log is active')` maybe?

tests/MaintenanceLockTest.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    /**
+     * @param $lockFileEnabled
+     * @param $fileExists
+     * @param $lockActive
+     * @dataProvider getMaintenanceCases
+     */
+    public function testEnableMaintenanceIfActive($lockFileEnabled, $fileExists, $lockActive)
+    {
+        Config::nest();
```
robbieaverill: Config nesting is done implicitly by SapphireTest in SS4 - you don't need this =)

tests/MaintenanceLockTest.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->assertFalse(file_exists($filePath));
+        $this->assertFalse(QueuedJobService::isMaintenanceLockActive());
+
+        Config::unnest();
+    }
+
+    /**
+     * @return array
+     */
+    public function getMaintenanceCases()
```
robbieaverill: Purely for convention this might be better named as `maintenanceCaseProvider`

src/Extensions/MaintenanceLockExtension.php
ScopeyNZ: I think this suggestion is functionally nicer as you're removing global state. With your suggestion you will be able to (theoretically) have two `QueuedJobService`s with different "maintenance lock" state.

mfendeksilverstripe:
@robbieaverill @ScopeyNZ New changes added, please review.

PR #267, opened 2019-11-09
FIX ignored "Start After" option when creating new Queued Job in CMS

JakubDolba:
Creating new Job in CMS Admin ignores Start at (after) field 
 - this cause problem in production: instead of delayed run, the job/task is executed almost immediately 

image


NightJar:
There is a linting error, and I'd like to check that this PR is targetting the correct branch before merging though - it's a patch, so probably should be at least onto the minor branch, rather than `master`.
But... sometimes there's this funny (peculiar) situation where the master is branch aliased as the minor branch.

NightJar:
Yes, thanks to @dnsl48 for tidying up the branches!
@JakubDolba could you please rebase to target either the 4.4 or 4.5 branch please (the former will be "merged up" to include the fix in 4.5 also).

PR #269, opened 2019-11-25
BUG Explictely register HTTPRequest with Injector in test setup

maxime-rainville:
Update the QueuedJobsAdminTest to explicitly register the fake HTTPRequest with the Injector.

This is causing problems for CWP Kitchen sink because some other LeftAndMain controller gets loaded and expects a valid request, but then it fails because it can't find one.

This should fix parts of the broken builds on the Kitchen sink for 2.2 and up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 env:
   global:
     - COMPOSER_ROOT_VERSION="4.3.x-dev"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,15 +1,17 @@
...
+
 env:
   global:
     - COMPOSER_ROOT_VERSION="4.3.x-dev"
 
 matrix:
   include:
     - php: 5.6
-      env: DB=MYSQL RECIPE_VERSION=1.0.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
+      env: DB=MYSQL RECIPE_VERSION=^1.0.0 PHPCS_TEST=1 PHPUNIT_TEST=1
```
dnsl48: Should we rather use the tilde in here `~` since caret would install 1.1

https://semver.mwl.be/#!?package=silverstripe%2Frecipe-cms&version=%5E1.0.0&minimum-stability=stable

.travis.yml
```diff
@@ -1,15 +1,17 @@
...
     - php: 7.0
-      env: DB=PGSQL RECIPE_VERSION=1.1.x-dev PHPUNIT_TEST=1
+      env: DB=PGSQL RECIPE_VERSION=^1.1.0 PHPUNIT_TEST=1
```
dnsl48: `~` would be more precise

.travis.yml
maxime-rainville: Good point. Just pushed a fix.

.travis.yml
maxime-rainville: Fixed.

dnsl48:
The tests are green. The linting issues are unrelated to this PR

PR #274, opened 2019-12-24
FIX: Make it possible to opt-out of "broken job" emails

### LINKED ISSUE #24: Prevent stalled job email spam
If the cron for queued jobs runs every minute, this means am email is sent every minute. It would be nice to only send an email e.g. once every hour. Especially if it's for the same stalled job.

Thanks!

lozcalver:
I’m opening this as a bugfix, rather than new feature, because I came into the office today to be greeted by 1,500 Sentry events logged, and that’s a bug in my opinion 😛

Closes #24

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -161,6 +161,15 @@ class QueuedJobService
      * @var string
      */
     private static $lock_file_path = '';
+    
+    /**
+     * Whether to log an error stating that there are broken jobs in the queue. This is usually
+     * used to send notification emails to developers/admins
+     *
+     * @config
+     * @var boolean
+     */
+    private static $notify_broken_jobs = true;
 
     /**
      * @var DefaultQueueHandler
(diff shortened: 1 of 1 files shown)


kinglozzer:
_pretty please_ 👼

dnsl48:
Why not use `disable_health_check` instead of adding a new flag?

dnsl48 (changes requested):
Could we try solving this problem by configuring the logger rather than adding more logging-related configs to the module?

I guess we could discuss that further in https://github.com/symbiote/silverstripe-queuedjobs/issues/24#issuecomment-596853647

chillu:
@mfendeksilverstripe Looks like you've added a datamodel for this with `NotifiedBroken`, but it doesn't appear to be used to reduce the errors. Can you clarify please? https://github.com/symbiote/silverstripe-queuedjobs/commit/4397484ebf1e22f4172ced6d86922cad73daf611

mfendeksilverstripe:
@chillu @kinglozzer I don't think [my job locking change-set](https://github.com/symbiote/silverstripe-queuedjobs/pull/279/files) made any changes that could increase the number of email reports on broken jobs.

I suspect that the change you are looking for is the [SS4 upgrade fix](https://github.com/symbiote/silverstripe-queuedjobs/pull/285/files). This fixes the email configuration and allows sending emails. I think that you were relying on broken functionality to not send emails about broken jobs and that fix enabled the broken jobs emails.

Happy to look further into the issues but please provide me with the following:

* commit hash of the version you are running
* if you see redundant broken job reports (re-occurring IDs)
* your yaml config for the queue

lozcalver:
Gah, I gave this PR a really stupid name... I wasn’t actually referring to the email notifications at all, I was actually referring to the error logging: I just was getting emails because they're logged to Sentry.

I think the new `NotifiedBroken` flag you added @mfendeksilverstripe should fix this (thanks!) as `checkJobHealth()` now seems to check that, I’ll try to update soon and build a test job that’ll definitely stall.

Either way I’ll close this PR, as it’s not a good solution. Will leave the issue #24 open for now until I (or someone else) has confirmed it’s fixed

PR #279, opened 2020-02-14
Thread safe job locking

mfendeksilverstripe:
# Thread safe job locking

This change-set introduces a thread safe job locking mechanism. It relies on Database features such as:

* transactions
* `SELECT FOR UPDATE`
* unique id generated for specific process

## Other changes

* multiple jobs are allowed to run at the same time by different processes as we can rely on job locking to prevent one jobs from being run more than once
* job run method uses now callback style for transactions, config nesting and log flushing
* `PHP 7.1` or higher is required
* each broken job is only reported once
* `JobErrorHandler` is now Injectible
* email related functionality moved to a separate service (no functionality changes)

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/280

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ dist: trusty
 
 matrix:
   include:
-    - php: 5.6
+    - php: 7.1
       env: DB=MYSQL RECIPE_VERSION=4.3.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
-    - php: 7.0
+    - php: 7.1
       env: DB=PGSQL RECIPE_VERSION=4.3.x-dev PHPUNIT_TEST=1
     - php: 7.1
       env: DB=MYSQL RECIPE_VERSION=4.4.x-dev PHPUNIT_TEST=1
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         }
     ],
     "require": {
+        "php": "^7.1",
         "silverstripe/framework": "^4",
         "silverstripe/admin": "^1",
         "asyncphp/doorman": "^3.0"
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -37,6 +37,10 @@ use Symbiote\QueuedJobs\Services\QueuedJobService;
  * @property string $SavedJobMessages List of messages saved for this job
  * @property string $JobStatus Status of this job
  * @property string $JobType Type of job
+ * @property string $Worker
+ * @property string $Expiry
+ * @property bool $NotifiedBroken
+ * @property int $WorkerCount
  *
  * @method Member RunAs() Member to run this job as
  *
@@ -70,6 +74,10 @@ class QueuedJobDescriptor extends DataObject
         'SavedJobMessages' => 'Text',
         'JobStatus' => 'Varchar(16)',
         'JobType' => 'Varchar(16)',
+        'Worker' => 'Varchar(32)',
+        'Expiry' => 'DBDatetime',
+        'NotifiedBroken' => 'Boolean',
+        'WorkerCount' => 'Int',
     ];
 
     /**
--- a/src/Services/EmailService.php
+++ b/src/Services/EmailService.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace Symbiote\QueuedJobs\Services;
+
+use SilverStripe\Control\Director;
+use SilverStripe\Control\Email\Email;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Subsites\Model\Subsite;
+
+/**
+ * Class EmailService
+ *
+ * Email reporting for queued jobs
+ *
+ * @package Symbiote\QueuedJobs\Services
+ */
+class EmailService
+{
+    use Injectable;
+
+    public function __construct()
+    {
+        $queuedEmail = Config::inst()->get(Email::class, 'queued_job_admin_email');
+
+        if ($queuedEmail || $queuedEmail === false) {
+            return;
+        }
+
+        // if not set (and not explictly set to false), fallback to the admin email.
+        Config::modify()->set(
+            Email::class,
+            'queued_job_admin_email',
+            Config::inst()->get(Email::class, 'admin_email')
+        );
+    }
+
+    /**
+     * @param array $jobConfig
+     * @param string $title
+     * @return Email|null
+     */
+    public function createMissingDefaultJobReport(array $jobConfig, string $title): ?Email
+    {
+        $subject = sprintf('Default Job "%s" missing', $title);
+        $from = Config::inst()->get('Email', 'queued_job_admin_email');
+        $to = array_key_exists('email', $jobConfig) &&  $jobConfig['email']
+            ? $jobConfig['email']
+            : $from;
+
+        if (!$to) {
+            return null;
+        }
+
+        return Email::create($from, $to, $subject)
+            ->setData($jobConfig)
+            ->addData('Title', $title)
+            ->addData('Site', Director::absoluteBaseURL())
+            ->setHTMLTemplate('QueuedJobsDefaultJob');
+    }
+
+    /**
+     * @param string $subject
+     * @param string $message
+     * @param int $jobID
+     * @return Email|null
+     */
+    public function createStalledJobReport(string $subject, string $message, int $jobID): ?Email
+    {
+        $email = $this->createReport($subject);
+        if ($email === null) {
+            return null;
+        }
+
+        return $email
+            ->setData([
+                'JobID' => $jobID,
+                'Message' => $message,
+                'Site' => Director::absoluteBaseURL(),
+            ])
+            ->setHTMLTemplate('QueuedJobsStalledJob');
+    }
+
+    /**
+     * Create a generic email report
+     * useful for reporting queue service issues
+     *
+     * @param string $subject
+     * @return Email|null
+     */
+    public function createReport(string $subject): ?Email
+    {
+        $from = Config::inst()->get(Email::class, 'admin_email');
+        $to = Config::inst()->get(Email::class, 'queued_job_admin_email');
+
+        if (!$to) {
+            return null;
+        }
+
+        return Email::create($from, $to, $subject);
+    }
+}
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -2,15 +2,17 @@
 
 namespace Symbiote\QueuedJobs\Services;
 
+use DateInterval;
+use DateTime;
 use Exception;
 use Monolog\Handler\BufferHandler;
 use Monolog\Logger;
 use Psr\Log\LoggerInterface;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
-use SilverStripe\Control\Email\Email;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Convert;
 use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
@@ -18,6 +20,8 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBDatetime;
+use SilverStripe\ORM\FieldType\DBField;
+use SilverStripe\ORM\Queries\SQLUpdate;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Security;
@@ -112,6 +116,14 @@ class QueuedJobService
     private static $max_init_jobs = 0;
 
     /**
+     * Duration for TTL of queue workers based on ISO 8601 duration specification.
+     * Defaults to 5 minutes.
+     *
+     * @var string
+     */
+    private static $worker_ttl = 'PT5M';
+
+    /**
      * Timestamp (in seconds) when the queue was started
      *
      * @var int
@@ -181,25 +193,17 @@ class QueuedJobService
     public $defaultJobs = [];
 
     /**
-     * Register our shutdown handler
+     * QueuedJobService constructor.
      */
     public function __construct()
     {
+        if (!$this->config()->get('use_shutdown_function') || !Director::is_cli()) {
+            return;
+        }
+
+        // Register our shutdown handler
         // bind a shutdown function to process all 'immediate' queued jobs if needed, but only in CLI mode
-        if (static::config()->get('use_shutdown_function') && Director::is_cli()) {
-            register_shutdown_function([$this, 'onShutdown']);
-        }
-
-        $queuedEmail = Config::inst()->get(Email::class, 'queued_job_admin_email');
-
-        // if not set (and not explictly set to false), fallback to the admin email.
-        if (!$queuedEmail && $queuedEmail !== false) {
-            Config::modify()->set(
-                Email::class,
-                'queued_job_admin_email',
-                Config::inst()->get(Email::class, 'admin_email')
-            );
-        }
+        register_shutdown_function([$this, 'onShutdown']);
     }
 
     /**
@@ -375,25 +379,18 @@ class QueuedJobService
         // Filter jobs by type
         $type = $type ?: QueuedJob::QUEUED;
         $list = QueuedJobDescriptor::get()
-            ->filter('JobType', $type)
+            ->filter([
+                'JobType' => $type,
+                'Worker:ExactMatch' => null,
+            ])
             ->sort('ID', 'ASC');
 
         // see if there's any blocked jobs that need to be resumed
         /** @var QueuedJobDescriptor $waitingJob */
-        $waitingJob = $list
-            ->filter('JobStatus', QueuedJob::STATUS_WAIT)
-            ->first();
+        $waitingJob = $list->find('JobStatus', QueuedJob::STATUS_WAIT);
+
         if ($waitingJob) {
             return $waitingJob;
-        }
-
-        // If there's an existing job either running or pending, the lets just return false to indicate
-        // that we're still executing
-        $runningJob = $list
-            ->filter('JobStatus', [QueuedJob::STATUS_INIT, QueuedJob::STATUS_RUN])
-            ->first();
-        if ($runningJob) {
-            return false;
         }
 
         // Otherwise, lets find any 'new' jobs that are waiting to execute
@@ -417,11 +414,16 @@ class QueuedJobService
      * handler or exception checker; in this case, we detect these stalled jobs later and fix (try) to
      * fix them
      *
+     * This function returns the IDs of stalled and broken jobs
+     * this information can be used to implement some custom follow up actions such as sending email reports
+     *
      * @param int $queue The queue to check against
+     * @return array stalled job and broken job IDs
      */
     public function checkJobHealth($queue = null)
     {
         $queue = $queue ?: QueuedJob::QUEUED;
+
         // Select all jobs currently marked as running
         $runningJobs = QueuedJobDescriptor::get()
             ->filter([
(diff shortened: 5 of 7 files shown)


robbieaverill:
Any thoughts on those test failures?

mfendeksilverstripe:
@robbieaverill I'll look into it ;-)

mfendeksilverstripe:
@robbieaverill Unit test are now green, please have another look

.travis.yml
```diff
@@ -4,10 +4,6 @@ dist: trusty
 
 matrix:
   include:
-    - php: 5.6
-      env: DB=MYSQL RECIPE_VERSION=4.3.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
-    - php: 7.0
-      env: DB=PGSQL RECIPE_VERSION=4.3.x-dev PHPUNIT_TEST=1
```
dnsl48, line 10: we should keep PostgreSQL within the test suite, maybe just update the PHP version here?

src/Services/QueuedJobService.php
```diff
@@ -434,42 +437,68 @@ public function checkJobHealth($queue = null)
 
         // If no steps have been processed since the last run, consider it a broken job
         // Only check jobs that have been viewed before. LastProcessedCount defaults to -1 on new jobs.
+        // Only check jobs that are past expiry to ensure another process isn't currently executing the job
+        $now = DBDatetime::now()->Rfc2822();
         $stalledJobs = $runningJobs
             ->filter('LastProcessedCount:GreaterThanOrEqual', 0)
-            ->where('"StepsProcessed" = "LastProcessedCount"');
+            ->where(sprintf('"StepsProcessed" = "LastProcessedCount" AND "Expiry" <= \'%s\'', $now));
```
dnsl48: Could we make it with bindings rather than concatenation?
I know, we're quite confident about what's within `$now` at this point, but it still would be nice to use bindings, especially that some databases may change sql query cache outcomes depending on that.

src/Services/QueuedJobService.php
```diff
@@ -434,42 +437,68 @@ public function checkJobHealth($queue = null)
...
+            ),
+            [
+                'file' => __FILE__,
+                'line' => __LINE__,
+            ]
+        );
+
+        DB::query(sprintf(
+            'UPDATE `QueuedJobDescriptor` SET `NotifiedBroken` = 1 WHERE `ID` IN (%s)',
+            implode(', ', $brokenIDs)
```
dnsl48: can we do this with bindings rather than concatenation?

src/Services/QueuedJobService.php
```diff
@@ -672,28 +703,59 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
-                'UPDATE "QueuedJobDescriptor" SET "JobStatus" = \'%s\' WHERE "ID" = %s'
-                . ' AND "JobFinished" IS NULL AND "JobStatus" NOT IN (%s)',
-                QueuedJob::STATUS_INIT,
-                $jobDescriptor->ID,
-                "'" . QueuedJob::STATUS_RUN . "', '" . QueuedJob::STATUS_COMPLETE . "', '"
-                . QueuedJob::STATUS_PAUSED . "', '" . QueuedJob::STATUS_CANCELLED . "'"
-            ));
+            // Start a transaction which will hold until we have a lock on this descriptor.
+            DB::get_conn()->withTransaction(function () use ($jobDescriptor) {
+                $query = 'SELECT `ID`, UUID() as `mutex`, `Worker` FROM `QueuedJobDescriptor`'
```
dnsl48: this is MySQL-specific syntax and won't work with other databases (e.g. PostgreSQL)

dnsl48: `UUID()` function is MySQL specific and won't work in other databases, can we generate it on PHP side?

src/Services/QueuedJobService.php
```diff
@@ -672,28 +703,59 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
+                    . ' WHERE `ID` = %s AND `Worker` IS NULL FOR UPDATE';
+
+                $row = DB::query(sprintf($query, $jobDescriptor->ID))->first();
```
dnsl48: We cannot use bindings here so we have to explicitly do `Convert::raw2sql` here.
That should either be `raw2sql((int) $id, false)`, or `raw2sql($id, true)`.

src/Services/QueuedJobService.php
```diff
@@ -672,28 +703,59 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
+                }
+
+                $mutex = $row['mutex'];
+                $workerExpiry = $this->getWorkerExpiry();
+
+                // Otherwise reserve this row.
+                $query = 'UPDATE `QueuedJobDescriptor` SET `Expiry` = \'%s\' , `Worker` = \'%s\','
+                    . ' `WorkerCount` = `WorkerCount` + 1 WHERE `ID` = %s';
+
+                DB::query(sprintf($query, $workerExpiry, $mutex, $jobDescriptor->ID));
```
dnsl48: This is MySQL specific without bindings, could we refactor please?

src/Services/QueuedJobService.php
```diff
@@ -672,28 +703,59 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
+
+                // Make sure we've got our row, and everything is fine.
+                $row = DB::query(
+                    sprintf('SELECT `ID`, `Worker` FROM `QueuedJobDescriptor` WHERE `ID` = %s', $jobDescriptor->ID)
```
dnsl48: Same here, bindings and ansi sql please (or better the ORM sql generator to solve both)

dnsl48 (changes requested):
Really well done! Thanks for the solid PR!
I haven't manually tested the patch yet, but skimmed it through and left a couple notes around about ANSI SQL compatibility. We still support PostgreSQL and shouldn't break it just yet.

emteknetnz:
Since you've added some protected functions, this will need to target `4` instead of `4.5`

mfendeksilverstripe:
@emteknetnz @dnsl48 Rebased on `4`, all unit tests are now green

composer.json
```diff
@@ -15,9 +15,11 @@
         }
     ],
     "require": {
+        "php": "^7.1",
         "silverstripe/framework": "^4",
         "silverstripe/admin": "^1",
-        "asyncphp/doorman": "^3.0"
+        "asyncphp/doorman": "^3.0",
+        "ramsey/uuid": "^3.9"
```
dnsl48: Any particular reason for using UUID specifically rather than simply generating a random id?
I think we can do without an extra 3rd party dependency here if we just go for something simpler, e.g.
```php
$mutex = base64_encode(random_bytes(32));
```

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -70,6 +74,10 @@ class QueuedJobDescriptor extends DataObject
         'SavedJobMessages' => 'Text',
         'JobStatus' => 'Varchar(16)',
         'JobType' => 'Varchar(16)',
+        'Worker' => 'Varchar(255)',
```
dnsl48: The `Worker` field would usually contain predictable number of characters, do we still need 255 here, or maybe we can define it more precisely?

src/Services/QueuedJobService.php
```diff
@@ -995,17 +1069,33 @@ public function runJob($jobId)
...
-        Config::unnest();
+        });
 
         $this->unsetRunAsUser($runAsUser, $originalUser);
 
         return !$broken;
     }
 
+    /**
+     * @param callable $callback
```
dnsl48, line 1056: Would be nice to have some documentation here. At least maybe some description of the main purpose of the method and the callback.

src/Services/QueuedJobService.php
```diff
@@ -995,17 +1069,33 @@ public function runJob($jobId)
...
+     * @return mixed
+     */
+    protected function withNestedState(callable $callback)
```
dnsl48, line 1059: Looks like a `static` method to me

src/Services/QueuedJobService.php
```diff
@@ -1363,6 +1453,47 @@ public function isMaintenanceLockActive()
         return file_exists($path);
     }
 
+    /**
+     * @param string $subject
+     * @param string|null $body
+     * @return Email|null
+     */
+    protected function generateEmailReport(string $subject, ?string $body = null): ?Email
```
dnsl48: can we move email functionality into a separate component? doesn't feel right to have it here.
Otherwise, could do with a private method instead?

src/Services/QueuedJobService.php
```diff
@@ -1363,6 +1453,47 @@ public function isMaintenanceLockActive()
         return file_exists($path);
     }
 
+    /**
+     * @param string $subject
```
dnsl48: some documentation would be nice

dnsl48 (changes requested):
Good work!
A couple final notes and suggestions from me

src/Services/QueuedJobService.php
mfendeksilverstripe: Definitely want to keep this `non-static` and `protected` method so it's possible to override with `Injector`.

mfendeksilverstripe:
@dnsl48 Pushed new changes, please review

src/Services/QueuedJobService.php
```diff
@@ -668,32 +685,66 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
+            DB::get_conn()->withTransaction(function () use ($descriptorId) {
+                $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "ID" = %s AND "Worker" IS NULL FOR UPDATE';
+
+                $row = DB::query(sprintf($query, Convert::raw2sql($descriptorId)))->first();
+
+                if (!array_key_exists('ID', $row) || !$row['ID']) {
+                    throw new Exception('Failed to read job lock');
+                }
+
+                $mutex = md5(sprintf('%s-%d-%d-%d', gethostname(), getmypid(), time(), mt_rand(1, 100000)));
```
dnsl48: `bin2hex(random_bytes(16))` would be more random and simpler to read and understand by other devs.
Also, it wouldn't utilise outdated and broken `md5` algo.
Since we obfuscate the original data by md5 anyways, do you reckon it might be worth refactoring? Speed and memory are negligible here, but simplicity might be worth pursuing.

src/Services/QueuedJobService.php
dnsl48: fair enough

dnsl48 (changes requested):
Looks cool overall, thank you for addressing everything!
One more note about `md5` usage and then I'd like to do some more manual testing before we merge. Otherwise, really well done, thank you!

src/Services/QueuedJobService.php
mfendeksilverstripe: Sounds good, thanks

mfendeksilverstripe:
@dnsl48 Updated the mutex generation, please review

src/Services/UnitTestLogger.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Class UnitTestLogger
+ *
+ * This logger is meant to be used only for Unit tests which produce unwanted console output
+ * this often breaks test suites like code coverage even though unit tests themselves pass
+ *
+ * @package Symbiote\QueuedJobs\Services
+ */
+class UnitTestLogger implements LoggerInterface, TestOnly
```
dnsl48: Can just use `TestCase::createMock(LoggerInterface::class)` instead. Let's remove this class.

tests/QueuedJobsTest.php
```diff
@@ -36,6 +42,18 @@ protected function setUp()
 
         // Allow large memory limit in cases of integration tests
         Config::modify()->set(QueuedJobService::class, 'memory_limit', 2 * 1024 * 1024 * 1024);
+
+        Injector::inst()->registerService(new UnitTestLogger(), LoggerInterface::class);
```
dnsl48:
```suggestion
        if (!Injector::inst()->has(LoggerInterface::class)) {
            Injector::inst()->registerService($this->createMock(LoggerInterface::class), LoggerInterface::class);
        }
```

src/Services/QueuedJobService.php
```diff
@@ -597,20 +619,15 @@ protected function restartStalledJob($stalledJob)
...
-        if ($to) {
-            $mail = Email::create($from, $to, $subject)
-                ->setData([
-                    'JobID' => $stalledJob->ID,
-                    'Message' => $message,
-                    'Site' => Director::absoluteBaseURL(),
-                ])
-                ->setHTMLTemplate('QueuedJobsStalledJob');
-            $mail->send();
+        if ($email === null) {
```
dnsl48:
```suggestion
        if ($email) {
            $email->send();
        }
```

dnsl48 (changes requested):
Tested locally, works like a charm. Thank you for the contribution!

Left a couple of the final small requests before we merge.

Also, could you please squash the commits and rewrite the message so it follows the [contribution guides](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages) so they get picked up by our changelog generator? (e.g. add NEW to the beginning of the messages). Ideally it would also have a paragraph explaining the feature.

mfendeksilverstripe:
@dnsl48 Implemented your feedback but I had to revert the `UnitTestLogger` removal as it breaks the test suites. [See more details here](https://travis-ci.org/symbiote/silverstripe-queuedjobs/builds/660326222?utm_source=github_status&utm_medium=notification).

What happens is that unit tests pass, but there is some additional console output which breaks the test suite.

I recommend reviewing the solution you proposed - I left it in a [separate commit](https://github.com/symbiote/silverstripe-queuedjobs/pull/279/commits/2463a6ea7f5b4c5bc7ba9b42d0c1fac114f0bca3).

Please let me know when this is ready to be squashed ;-)

dnsl48:
>  I had to revert the UnitTestLogger removal as it breaks the test suites

that's interesting. I was testing that fix locally and tests were passing for me.
I'll investigate further what exactly goes wrong on Travis.

P.S. looks like it's only happening in 4.3 (IIRC we changed some logger settings in 4.4). Might actually just remove 4.3 from the test suite as it's out of support anyways and this PR targeting the next minor of the module anyway. But let me have a look at it first.

dnsl48:
@mfendeksilverstripe 
I've had some debugging and 4.3 seems to be having a default logger preset in there, so `Injector::inst()->has()` returns a default logger in 4.3 and nothing in >=4.4.

I reckon we should just replace the initial logger without checking for `has()`.
The original logger gonna be restored by SapphireTest tearDown for the further tests anyways.

Either way, the problem is not with mocking the interface.

robbieaverill:
@mfendeksilverstripe @dnsl48 I can help a bit there - in 4.4 we changed logging so that LoggerInterface is entirely a user defined logger. Previously it was used for the core error handling. Since 4.4 it's `LoggerInterface::class . '. errorhandler'` that does core stuff. See https://github.com/silverstripe/silverstripe-framework/pull/8241 for context.

mfendeksilverstripe:
@dnsl48 I applied your fix, squashed and updated commit message. Test are green, all is packaged and good to go ;-)

dnsl48 (approved):
Well done, thank you!

src/Services/EmailService.php
```diff
@@ -0,0 +1,102 @@
...
+    public function __construct()
+    {
+        $queuedEmail = Config::inst()->get(Email::class, 'queued_job_admin_email');
+
+        if ($queuedEmail || $queuedEmail === false) {
+            return;
+        }
+
+        // if not set (and not explictly set to false), fallback to the admin email.
+        Config::modify()->set(
```
robbieaverill, line 31: Mutating the config manifest does have a slight performance hit (if it hasn't been mutated already in the request). It might be slightly more performant if you were to put this fallback logic into the `createReport()` method, e.g.:

```php
$to = Config::inst()->get(Email::class, 'queued_job_admin_email') ?: Config::inst()->get(Email::class, 'admin_email');
```

_Edit:_ I realise you've moved this from existing code

src/Services/EmailService.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @param array $jobConfig
+     * @param string $title
+     * @return Email|null
+     */
+    public function createMissingDefaultJobReport(array $jobConfig, string $title): ?Email
+    {
+        $subject = sprintf('Default Job "%s" missing', $title);
+        $from = Config::inst()->get('Email', 'queued_job_admin_email');
```
robbieaverill, line 46:
```suggestion
        $from = Config::inst()->get(Email::class, 'queued_job_admin_email');
```

src/Services/QueuedJobService.php
```diff
@@ -597,19 +619,12 @@ protected function restartStalledJob($stalledJob)
                 'line' => __LINE__,
             ]
         );
-        $from = Config::inst()->get(Email::class, 'admin_email');
-        $to = Config::inst()->get(Email::class, 'queued_job_admin_email');
+
         $subject = _t(__CLASS__ . '.STALLED_JOB', 'Stalled job');
+        $email = EmailService::singleton()->createStalledJobReport($subject, $message, (int) $stalledJob->ID);
```
robbieaverill, line 624: You could use `private static $dependencies` to inject `EmailService` if you wanted to

src/Services/QueuedJobService.php
```diff
@@ -668,32 +683,66 @@ protected function initialiseJob(QueuedJobDescriptor $jobDescriptor)
...
-                "'" . QueuedJob::STATUS_RUN . "', '" . QueuedJob::STATUS_COMPLETE . "', '"
-                . QueuedJob::STATUS_PAUSED . "', '" . QueuedJob::STATUS_CANCELLED . "'"
-            ));
+            // Start a transaction which will hold until we have a lock on this descriptor.
+            DB::get_conn()->withTransaction(function () use ($descriptorId) {
+                $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "ID" = %s AND "Worker" IS NULL FOR UPDATE';
+
+                $row = DB::query(sprintf($query, Convert::raw2sql($descriptorId)))->first();
+
+                if (!array_key_exists('ID', $row) || !$row['ID']) {
```
robbieaverill, line 699:
```suggestion
                if (empty($row['ID'])) {
```

src/Services/QueuedJobService.php
```diff
@@ -1363,6 +1429,29 @@ public function isMaintenanceLockActive()
...
+     * @throws Exception
+     */
+    protected function getWorkerExpiry(): string
+    {
+        $now = DBDatetime::now()->Rfc2822();
+        $time = new DateTime($now);
+        $timeToLive = $this->config()->get('worker_ttl');
+
+        if ($timeToLive) {
+            $time->add(new DateInterval($timeToLive));
```
robbieaverill, line 1446: You can actually do this with only DBDatetime ([reference](https://github.com/silverstripe/silverstripe-framework/pull/9105/files)), example:

```php
$expiry = DBDatetime::now();
$timeToLive = $this->config()->get('worker_ttl'); // example: "+5 minutes"
if ($timeToLive) {
    $expiry->modify($timeToLive);
}
return $expiry->Rfc2822();
```

robbieaverill (commented):
Sorry for the late review, I've left a couple of suggestions

src/Services/EmailService.php
mfendeksilverstripe: correct, there are no functional changes here

src/Services/QueuedJobService.php
mfendeksilverstripe: That's good to know, thanks :)

dnsl48:
Sorry @robbieaverill, I merged too quick

robbieaverill:
No no I should've reviewed it sooner 😛

src/Services/QueuedJobService.php
mfendeksilverstripe: We actually have a litning rule set up to avoid using `empty`. Maybe it's ok in this situation, though

src/Services/EmailService.php
mfendeksilverstripe: Who, this looks like a SS4 upgrade issue, nice catch :)

PR #283, opened 2020-02-26
Document required fields for defaultJobs

oddnoc:
There are three fields required in the code that were not documented as REQUIRED for defaultJobs. The code is at QueuedJobService#checkDefaultJobs(), line 533.

* construct (incorrectly documented as OPTIONAL)
* startDateFormat (not documented)
* startTimeString (not documented)


dnsl48 (approved):
Thank you for the contribution!

PR #284, opened 2020-03-09
Fix QueuedJobsAdmin parsing job parameters

dnsl48:
QueuedJobsAdmin would try to split parameters by PHP_EOL,
which is a single newline (LF aka "\n") on unix based systems.
That causes incorrect argument parsing when browsers send
CRLF ("\r\n").

Fixes https://github.com/symbiote/silverstripe-queuedjobs/issues/258


Cheddam (approved):
Classic! Nice work :)

PR #285, opened 2020-03-10
BUG SilverStripe 4 upgrade fix

mfendeksilverstripe:
This fixes an outdated class reference which was missed during SilverStripe 4 upgrade

This issue was detected but not caused in https://github.com/symbiote/silverstripe-queuedjobs/pull/279

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/286


Cheddam (approved):
Nice catch 😄

PR #263, opened 2019-09-20
Exception handling fix (on broken setup()).

mfendeksilverstripe:
Fixed exception handling in the case of setup() function ended up throwing.

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/276

PR #289, opened 2020-04-16
BUG: Broken job handling when setup() throws

mfendeksilverstripe:
Apply the https://github.com/symbiote/silverstripe-queuedjobs/pull/263 to `4`

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -1077,7 +1077,7 @@ class QueuedJobService
      * @param QueuedJob|null $job
      * @param Exception|\Throwable $e
      */
-    protected function handleBrokenJobException(QueuedJobDescriptor $jobDescriptor, $job, $e)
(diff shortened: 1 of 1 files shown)


mfendeksilverstripe:
@Cheddam Could you please have a look at this one?

Cheddam:
@mfendeksilverstripe I'm a little confused - you used PHP 7.1-compatible syntax in #263 to make the `$job` parameter nullable, but here you drop the typehint entirely?

mfendeksilverstripe:
You're right @Cheddam , for some reason I thought that nullable type hint is only since PHP `7.2`. I updated the PR, please review.

Cheddam:
Thanks for the quick turnaround! Regrettably, after further review, this constitutes a major change. It'd be fine if `$job` was only used internally in this method, but @dnsl48 noted that it changes the signature of the `updateJobDescriptorAndJobOnException` extension point, which could break downstream code. On that basis, I don't think we can bring this fix to the `4` branch, sorry Mojmir 😞

mfendeksilverstripe:
Fair enough, thanks for your help anyway

PR #293, opened 2020-05-21
BUGFIX: ['stalled'] needs to be defined to print values

mattclegg:
`$this->getService()->checkJobHealth($queue);` returns an array which  means `echo $stalledJobCount === 0` fails. I have updated this to show some logical output and also added an exit(1) for when this is run from cli to detect unhealthy jobs.

Any feedback welcome.

### DIFF
--- a/src/Tasks/CheckJobHealthTask.php
+++ b/src/Tasks/CheckJobHealthTask.php
@@ -2,6 +2,7 @@
 
 namespace Symbiote\QueuedJobs\Tasks;
 
+use Exception;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Dev\BuildTask;
 use Symbiote\QueuedJobs\Services\QueuedJob;
@@ -34,6 +35,8 @@ class CheckJobHealthTask extends BuildTask
      *
      * @param HTTPRequest $request
      * @return
+     *
+     * @throws Exception
      */
     public function run($request)
     {
@@ -48,11 +51,11 @@ class CheckJobHealthTask extends BuildTask
             $unhealthyJobCount = $unhealthyJobCount + $count;
         }
 
-        if ($unhealthyJobCount === 0) {
-            echo 'All jobs are healthy';
-        } else {
-            die(1);
+        if ($unhealthyJobCount > 0) {
+            throw new Exception("$unhealthyJobCount jobs are unhealthy");
         }
+
+        echo 'All jobs are healthy';
     }
 
     protected function getService()
--- a/src/Tasks/CheckJobHealthTask.php
+++ b/src/Tasks/CheckJobHealthTask.php
@@ -38,11 +38,21 @@ class CheckJobHealthTask extends BuildTask
     public function run($request)
     {
         $queue = $request->requestVar('queue') ?: QueuedJob::QUEUED;
+        $jobHealth = $this->getService()->checkJobHealth($queue);
 
-        $stalledJobCount = $this->getService()->checkJobHealth($queue);
+        $unhealthyJobCount = 0;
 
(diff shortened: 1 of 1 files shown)


src/Tasks/CheckJobHealthTask.php
```diff
@@ -38,11 +38,21 @@ public function getDescription()
...
+        foreach ( $jobHealth as $type => $IDs ) {
+            $count = count($IDs);
+            echo 'Detected and attempted restart on ' . $count . ' ' . $type . ' jobs';
+            $unhealthyJobCount = $unhealthyJobCount + $count;
+        }
+
+        if( $unhealthyJobCount === 0 ) {
+            echo 'All jobs are healthy';
+        } else {
+            die(1);
```
emteknetnz: Are you sure we'd want to `die(1)` here? Previous implementation did not die

mattclegg: This sets the exit code to be '1'. This would be useful for CLI implementations so you can tell if `All jobs are healthy` from the exit code.

emteknetnz: OK makes sense. What happens if you run this from a browser?

mattclegg: PHP will halt execution, but as it's the end of the Task then it wont appear any different to the browser.

mattclegg: It's worth noting that die & exit are also equivalent.

Ref; https://www.php.net/manual/en/function.exit.php#refsect1-function.exit-examples

dhensby: At the moment there is no die, so this is new behaviour isn't it? `die` will also halt all post-request execution, which might be required. I don't think we should start adding `exit`/`die` calls if they aren't there already...

We could throw an exception which I believe will give an error code on CLI and would allow the post-execution scripts to run.

dhensby (changes requested):

src/Tasks/CheckJobHealthTask.php
```diff
@@ -48,11 +48,11 @@ public function run($request)
             $unhealthyJobCount = $unhealthyJobCount + $count;
         }
 
-        if ($unhealthyJobCount === 0) {
-            echo 'All jobs are healthy';
-        } else {
-            die(1);
+        if ($unhealthyJobCount > 0) {
+            throw new Exception("$unhealthyJobCount jobs are unhealthy");
```
dhensby, line 55: How about a more specific exception? We could create an exception like "UnhealthJobsException".

dhensby (changes requested):

src/Tasks/CheckJobHealthTask.php
mattclegg: True, I've updated it. Although throwing an exception will also halt all post-request execution?

src/Tasks/CheckJobHealthTask.php
mattclegg: Maybe that should be another PR as the module is still using `Exception` in 5 other places?

```
$ grep "use Exception;" .
./tests/QueuedJobsTest/TestExceptingJob.php:7:use Exception;
./src/Tasks/PublishItemsTask.php:5:use Exception;
./src/Jobs/GenerateGoogleSitemapJob.php:5:use Exception;
./src/Services/QueuedJobService.php:7:use Exception;
./src/Services/JobErrorHandler.php:5:use Exception;
```

src/Tasks/CheckJobHealthTask.php
dhensby: It won't because all the catch / finally blocks will still run

dhensby (approved):

PR #294, opened 2020-05-27
Check descriptor exists to handle cases when it is deleted.

wilr:
Handle cases better when `refreshDescriptor` is called and no descriptor is found.

### DIFF
--- a/code/jobs/DoormanQueuedJobTask.php
+++ b/code/jobs/DoormanQueuedJobTask.php
@@ -6,185 +6,207 @@ use AsyncPHP\Doorman\Process;
 use AsyncPHP\Doorman\Task;
 
 class DoormanQueuedJobTask implements Task, Expires, Process, Cancellable {
-	/**
-	 * @var int
-	 */
-	protected $id;
+    /**
+     * @var int
+     */
+    protected $id;
 
-	/**
-	 * @var QueuedJobDescriptor
-	 */
-	protected $descriptor;
+    /**
+     * @var QueuedJobDescriptor
+     */
+    protected $descriptor;
 
-	/**
-	 * Reload descriptor from DB
-	 */
-	protected function refreshDescriptor() {
-		if($this->descriptor) {
-			$this->descriptor = QueuedJobDescriptor::get()->byID($this->descriptor->ID);
-		}
-	}
+    /**
+     * Reload descriptor from DB
+     */
+    protected function refreshDescriptor() {
+        if ($this->descriptor) {
+            $this->descriptor = QueuedJobDescriptor::get()->byID($this->descriptor->ID);
+        }
+    }
 
-	/**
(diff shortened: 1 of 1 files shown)


code/jobs/DoormanQueuedJobTask.php
```diff
@@ -185,6 +200,13 @@ public function canRunTask() {
...
 		$this->refreshDescriptor();
-		return $this->descriptor->JobStatus === QueuedJob::STATUS_CANCELLED;
+
+		$descriptor = $this->getDescriptor();
+
+		if ($descriptor) {
+			return $this->descriptor->JobStatus === QueuedJob::STATUS_CANCELLED;
+		}
+
+		return true;
```
emteknetnz: This should be `return false`?

Missing descriptor doesn't really mean it was cancelled, just that it's missing?

code/jobs/DoormanQueuedJobTask.php
```diff
@@ -168,14 +176,21 @@ public function shouldExpire($startedAt) {
...
-				QueuedJob::STATUS_WAIT
-			)
-		);
+
+		$descriptor = $this->getDescriptor();
+
+		if ($descriptor) {
+			return in_array(
+				$descriptor->JobStatus,
+				array(
```
emteknetnz: should use short array syntax

emteknetnz (changes requested):
Looks good, question around the 'default' value of isCancelled()

code/jobs/DoormanQueuedJobTask.php
wilr: Changed to false

code/jobs/DoormanQueuedJobTask.php
wilr: Updated to short syntax

wilr:
@emteknetnz Updated based on feedback

emteknetnz:
Great ta, I'll test this locally later today.

emteknetnz:
Apologies, I didn't realise this was for Silverstripe 3 instead of Silverstripe 4

I probably gave bad advice re the short-array syntax, since 3.1 of this module requires framework ^3.1 it technically needs to support php 5.3 which does not support short array syntax =(

Technically we (Silverstripe Ltd who I work for) only have limited support for Silverstripe 3 issues meaning only critical and security related.  This pull-request isn't covered by that.  See https://www.silverstripe.org/software/roadmap/.

I guess I can still approve this one for you though based on code-review (it looks fine), though testing/merging will need to be a community thing

Sorry once again =/

GuySartorelli:
The branch this is targeting is no longer supported.
If this affects Silverstripe CMS 5, please target the `5.1` branch.

wilr:
Looks like it has been fixed independently. this->describe is checked before the call.

PR #301, opened 2020-06-08
NEW: Task runner better UI

mfendeksilverstripe:
# Task runner better UI

Simple UI changes which improve usability of the task runner. Lightweight CSS (no build chain or JS required).

## Problem

* it takes time to find the dev task you want to run especially if you don't know the exact task name
* links which execute dev tasks are tiny, mis-click error can happen
* current UI is cluttered with too many dev tasks and the space is used inefficiently
* rendering component of the task runner is messy (combines backend and frontend logic)
* overriding frontend display requires subclassing

### Current UI
Screen Shot 2020-06-08 at 12 05 05 PM

## Solution

* move the task groups into separate tabs so the screen is less cluttered
* visually separate the dev tasks
* visually separate the action buttons and make them much more visible
* split the rendering into backend (populate data) and a template (render), this allows the template to be overridden without subclassing

### Proposed UI
Screen Shot 2020-06-08 at 1 16 11 PM

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/302

## Dependencies

https://github.com/silverstripe/silverstripe-framework/pull/9540

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,8 +10,9 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[{*.yml,package.json}]
-indent_size = 2
+[{*.yml,*.ss,*.css,*.json}]
+indent_size = 4
+indent_style = space
 
 # The indent size used in the package.json file cannot be changed:
 # https://github.com/npm/npm/pull/3180#issuecomment-16336516
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,9 +10,22 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[{*.yml,*.ss,*.css,*.json}]
-indent_size = 4
+[*.md]
+trim_trailing_whitespace = false
+
+[*.{yml,js,json,css,scss,eslintrc,feature}]
+indent_size = 2
 indent_style = space
 
-# The indent size used in the package.json file cannot be changed:
-# https://github.com/npm/npm/pull/3180#issuecomment-16336516
+[composer.json]
+indent_size = 4
+
+# Don't perform any clean-up on thirdparty files
+
+[thirdparty/**]
+trim_trailing_whitespace = false
+insert_final_newline = false
+
+[admin/thirdparty/**]
+trim_trailing_whitespace = false
+insert_final_newline = false
--- a/client/styles/task-runner.css
+++ b/client/styles/task-runner.css
@@ -11,20 +11,8 @@
 .task__selector {
     display: inline-block;
     height: 42px;
-    margin: 0 0 0 15px;
+    margin: 0;
     visibility: hidden;
-}
-
-.task__selector--universal {
-  width: 105px;
-}
-
-.task__selector--immediate {
-  width: 144px;
-}
-
-.task__selector--queue-only {
-  width: 146px;
 }
 
 .task__selector:checked + .task__label .task__label-inner,
@@ -37,21 +25,6 @@
     transform: translateY(-15px);
 }
 
-.task__label--universal {
-  margin-left: -109px;
-  width: 105px;
-}
-
-.task__label--immediate {
-  margin-left: -147px;
-  width: 144px;
-}
-
-.task__label--queue-only {
-  margin-left: -150px;
-  width: 146px;
-}
-
 .task__label-inner {
   border-bottom: 2px solid transparent;
   padding-bottom: 15px;
@@ -62,12 +35,13 @@
     margin-left: 0;
 }
 
-.task__panel {
+.task__panel .task__item {
     display: none;
 }
 
-.task__selector--immediate:checked ~ .task__panel--immediate,
-.task__selector--universal:checked ~ .task__panel--universal,
-.task__selector--queue-only:checked ~ .task__panel--queue-only {
-    display: block;
+.task__selector--immediate:checked ~ .task__panel .task__item--immediate,
+.task__selector--universal:checked ~ .task__panel .task__item--universal,
+.task__selector--queue-only:checked ~ .task__panel .task__item--queue-only,
+.task__selector--all:checked ~ .task__panel .task__item {
+    display: inline-block;
 }
--- a/client/styles/task-runner.css
+++ b/client/styles/task-runner.css
@@ -54,5 +54,5 @@
 }
 
 .task__button + .task__button {
-    margin-left: 25px;
+    margin-left: 5px;
 }
--- a/client/styles/task-runner.css
+++ b/client/styles/task-runner.css
@@ -29,7 +29,7 @@
 
 .task__selector:checked + .task__label .task__label-inner,
 .task__selector:hover + .task__label .task__label-inner {
-  border-bottom-color: #475266;
+  border-bottom-color: #43536d;
 }
 
 .task__label {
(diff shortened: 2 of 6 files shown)


.editorconfig
```diff
@@ -10,8 +10,9 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[{*.yml,package.json}]
+[{*.yml,*.ss,*.css,*.json}]
```
michalkleiner: Just make it 4 spaces everywhere... ;-)

mfendeksilverstripe: Done and done :)

michalkleiner: I expect some pushback from the core team as it's not a part of the functional change, but tbh I never understood the need for 2 spaces for `.yml` and `.json` only.

michalkleiner:
Very nice touch! Often these sparingly used areas of the CMS/admin interface don't get any updates/tweaks, so this is refreshing to see.

I'd suggest looking at it at the framework level though, if possible, where it lives without the queued jobs additions (run now/enqueue). Here it can subsequently alter the actions.

mfendeksilverstripe:
Will have a look at the core task runner. Thanks for feedback @michalkleiner

mfendeksilverstripe:
@michalkleiner Moved some of the code to the framework, please review.

.editorconfig
emteknetnz: Just copy whatever is in https://github.com/silverstripe/silverstripe-framework/blob/4/.editorconfig

> the need for 2 spaces

I don't know either, though consistent standards mean we don't need to think or spend time debating stuff =)

mfendeksilverstripe: @emteknetnz  Editor config updated from framework

sachajudd (changes requested):
Review comments for framework can be found here but also apply to this PR: https://github.com/silverstripe/silverstripe-framework/pull/9540#pullrequestreview-441332266

mfendeksilverstripe:
@sachajudd Pushed up new changes based on UI feedback, please review. This is how the UI looks like now:

## Large screen
Screen Shot 2020-07-17 at 11 37 53 AM

## Small screen
Screen Shot 2020-07-17 at 11 38 22 AM

client/styles/task-runner.css
```diff
@@ -0,0 +1,73 @@
...
+  width: 144px;
+}
+
+.task__selector--queue-only {
+  width: 146px;
+}
+
+.task__selector:checked + .task__label .task__label-inner,
+.task__selector:hover + .task__label .task__label-inner {
+  border-bottom-color: #475266;
```
sachajudd: What's this colour for? Should it be #43536d?

client/styles/task-runner.css
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.task__label--queue-only {
+  margin-left: -150px;
+  width: 146px;
+}
+
+.task__label-inner {
+  border-bottom: 2px solid transparent;
+  padding-bottom: 10px;
```
sachajudd: Looks a bit tight in the UI, try 15px

sachajudd:
See comments on https://github.com/silverstripe/silverstripe-framework/pull/9540#issuecomment-659760122 which apply to this PR 🙂

client/styles/task-runner.css
mfendeksilverstripe: good catch, thanks :)

mfendeksilverstripe:
New changes:

- border hover colour fixed
- more space for bottom border

mfendeksilverstripe:
New changes:

- updates labels in other tabs.

This is now ready for a review.

sachajudd (approved):
Approved from a UX perspective 🙂

client/styles/task-runner.css
```diff
@@ -0,0 +1,73 @@
...
+
+.task__panel {
+    display: none;
+}
+
+.task__selector--immediate:checked ~ .task__panel--immediate,
+.task__selector--universal:checked ~ .task__panel--universal,
+.task__selector--queue-only:checked ~ .task__panel--queue-only {
+    display: block;
+}
```
Cheddam: Love the technique - and I think you could take it a step further. At the moment you're grouping the tasks into separate sections, which can result in some inconsistent spacing between items at the split. Instead, you could group them back together, shift the modifier to the `.task__item` level, and use an adjusted selector to hide / show individual tasks:

```suggestion
.task__selector--immediate:checked ~ .task__panel .task__item--immediate,
.task__selector--universal:checked ~ .task__panel .task__item--universal,
.task__selector--queue-only:checked ~ .task__panel .task__item--queue-only {
    display: block;
}
```

This would also mean you could reduce duplication in the template, and instead have conditionals around the task buttons.

Of course, keeping the types split into separate sections might be desirable. If this is the case, I'd at least suggest adding a heading to each section to make this separation clearer.

templates/Symbiote/QueuedJobs/Controllers/QueuedTaskRunner.ss
```diff
@@ -0,0 +1,76 @@
+$Header.RAW
+$Info.RAW
+
+<div class="task">
+    <input type="radio" id="tab-universal-tasks" class="task__selector task__selector--universal" name="task__selector" checked="checked" />
+    <label class="task__label task__label--universal" for="tab-universal-tasks">
+        <span class="task__label-inner">Queueable task</span>
```
Cheddam, line 12: I think the default tab should make all tasks visible, rather than only 'universal' tasks. That might mean having four tabs if you want to keep this filter.

client/styles/task-runner.css
```diff
@@ -0,0 +1,73 @@
...
+
+.task__label--immediate {
+  margin-left: -147px;
+  width: 144px;
+}
+
+.task__label--queue-only {
+  margin-left: -150px;
+  width: 146px;
+}
```
Cheddam: I don't think these rules (along with the hardcoded widths on the `input`s) are necessary to align / size these elements. I get almost identical output (pushed to the right by a couple of pixels) by dropping both of these sets of selectors out, and changing the margin on `.task__selector` to `0`.

Cheddam (changes requested):
Hey Mojmir! This PR, along with the core one, is a great step forward for the dev/tasks UI. I've suggested some fairly minor modifications.

I'm also pondering whether it'd be sensible to reduce the code duplication between the base TaskRunner and this one by splitting some of the functionality out into separate methods / config? For example, the stylesheets could be stored in configuration, and the logic for handling them could be split into a method on the base TaskRunner class. This might be overkill, but it might also serve to reduce the amount of double-handling required in future updates to this UI (for example, I'm keen to introduce proper categorisation of tasks later on, along with a search field.)

Let me know what you think.

client/styles/task-runner.css
mfendeksilverstripe: Great suggestion, thanks :)

client/styles/task-runner.css
mfendeksilverstripe: Nice catch, let's remove it :)

mfendeksilverstripe:
@Cheddam Pushed up changes as requested, please review.

mfendeksilverstripe:
@Cheddam Pushed up changes as requested, please review.

Cheddam (approved):
Excellent work, thanks for your contribution!

Cheddam (changes requested):
Actually, one last minor tweak is required 😅 

Since we standardised the way the CSS is included in the framework PR, this PR now needs to bump the minimum framework requirement to `^4.7` (the next minor release). I'll tweak this now.

Cheddam (approved):
Okay, I've also rebased onto the `4` branch since this is a new feature, and fixed up Travis config. Once the tests go green we're good to merge 👍

PR #305, opened 2020-06-09
BUG: Doorman manager child processing fix.

mfendeksilverstripe:
# Doorman manager child processing fix

## Problem
Using asynchronous queue runner has a flaw - when the manager process terminates, it also terminates all of its child processes. This is a mistake because there is a valid use case where manager process doesn't have any more work to do (no more jobs to start) but the child processes are still executing jobs.

## Solution
Fix the shell command and prevent the manager process from killing of the child processes.

## Notes

* Requires 3.1.0 version or higher of `asyncphp/doorman`
* split from [Feature batch](https://github.com/symbiote/silverstripe-queuedjobs/pull/290) as `Feature 9 - Job manager process`

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/306

### DIFF
--- a/src/Services/ProcessManager.php
+++ b/src/Services/ProcessManager.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Symbiote\QueuedJobs\Services;
+
+use AsyncPHP\Doorman\Manager\ProcessManager as BaseProcessManager;
+
+/**
+ * Class ProcessManager
+ *
+ * customise shell command to allow child tasks to persist even after manager process is terminated
+ * this lets the started jobs to finish properly in case the management process terminates
+ * fore example there are no more jobs to start or queue is paused
+ *
+ * @package Symbiote\QueuedJobs\Services
+ */
+class ProcessManager extends BaseProcessManager
+{
+    /**
+     * @param string $binary
+     * @param string $worker
+     * @param string $stdout
+     * @param string $stderr
+     * @return string
+     */
+    protected function getCommand($binary, $worker, $stdout, $stderr) // phpcs:ignore SlevomatCodingStandard.TypeHints
+    {
+        return sprintf('nohup %s %s %%s %s %s & echo $!', $binary, $worker, $stdout, $stderr);
+    }
+
+    public function __destruct()
+    {
+        // Prevent background tasks from being killed when this script finishes
+        // this is an override for the default behaviour of killing background tasks
+    }
+}
--- a/src/Services/ProcessManager.php
+++ b/src/Services/ProcessManager.php
@@ -3,6 +3,7 @@
 namespace Symbiote\QueuedJobs\Services;
 
 use AsyncPHP\Doorman\Manager\ProcessManager as BaseProcessManager;
+use SilverStripe\Core\Config\Configurable;
 
 /**
  * Class ProcessManager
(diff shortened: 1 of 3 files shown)


src/Services/ProcessManager.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * @param string $binary
+     * @param string $worker
+     * @param string $stdout
+     * @param string $stderr
+     * @return string
+     */
+    protected function getCommand($binary, $worker, $stdout, $stderr) // phpcs:ignore SlevomatCodingStandard.TypeHints
+    {
+        return sprintf('nohup %s %s %%s %s %s & echo $!', $binary, $worker, $stdout, $stderr);
```
emteknetnz, line 54: Is there any chance the server won't have `nohup` installed?  Is it worth testing for `which nohup` first?

src/Services/ProcessManager.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function __destruct()
+    {
+        // Prevent background tasks from being killed when this script finishes
```
emteknetnz: Is there a valid reason why the default behaviour may be good? Is it worth making this behaviour a configurable value?

emteknetnz (changes requested):
Code wise its fine, though a couple of questions

You'll need to rebase/retarget this to `4` since it has new API

Also, would it be worth simply making this PR straight to asyncphp/doorman instead?

src/Services/ProcessManager.php
dhensby, line 54: I think this needs some commenting about what is going on here (for those less familiar with `nohup` and `&` and `$!` in bash

mfendeksilverstripe:
Rebased to `4`, pushed up new changes, please review.

src/Services/ProcessManager.php
mfendeksilverstripe: I don't think we need to cover this edge case. If there is such situation there is an easy option to subclass and override.

src/Services/ProcessManager.php
mfendeksilverstripe: The default behaviour is generally considered a more safe option and I would still want to keep it as a default. The only reason why we can bypass this here is that this module has a very good system in place which keeps track of running jobs (job locking...).

mfendeksilverstripe:
@emteknetnz To answer your question:

> Also, would it be worth simply making this PR straight to asyncphp/doorman instead?

I don't think this changeset belongs to the doorman module as the change by itself is probably a bit dangerous default. In the context of this module it's different as there are other mechanisms in place which makes this change reasonable (job locking, execution timeouts).

src/Services/ProcessManager.php
emteknetnz: Ok, so you're saying that it's no worth making this configurable because the queued jobs module does such a good job?
Should we do either of the following?
a) make this configurable?
b) add your rationale for why it's ok to bypass into the inline comments?

emteknetnz:
Mojmir, looking good, just a couple of questions around the destruct function

src/Services/ProcessManager.php
mfendeksilverstripe: I had a look at it in detail and I think it's worth making it configurable. There might be some cases where you want this to be disabled.

mfendeksilverstripe:
@emteknetnz Pushed up new changes (configurable child process termination + docs), please review.

dhensby (approved):

PR #312, opened 2020-06-10
NEW: Show job data in descriptor edit form.

mfendeksilverstripe:
# Show job data in descriptor edit form

Add the option to show read only version of raw job data and raw messages in the job descriptor edit form. Option is disabled by default.

## Problem
Job data is inaccessible in the jobs admin which makes debugging jobs quite difficult. It often requires the developer to download DB snapshot in order to do anything. In case job messages get corrupted (for example in case of getting truncated), they become inaccessible forcing the developer to rely on DB dump instead.

## Solution
Expose job data and raw messages as read only fields in the job edit form.

Screen Shot 2020-06-11 at 9 17 55 AM

Screen Shot 2020-06-11 at 9 18 02 AM

## Notes

* split from [Feature batch](https://github.com/symbiote/silverstripe-queuedjobs/pull/290) as Feature 2 - Improved Job edit form

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/313

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,6 +819,17 @@ As a consequence, the work might end up being very fragmented and each chunk may
 
 Some projects do not mind this however, so this solution may still be quite suitable.
 
+## Show job data
+
+In case you need an easy access to additonal job data via CMS for debug purposes enable the `show_job_data` option by including the configuration below.
+
+```yaml
+Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor:
+  show_job_data: true
+```
+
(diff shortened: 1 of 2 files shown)


emteknetnz:
@mfendeksilverstripe this one will need to target `4` as it has new public functions / API

Code looks seems fine otherwise

emteknetnz (changes requested):
Look fine, you'll need to rebase/retarget this to `4` since it contains new API

mfendeksilverstripe:
@emteknetnz Rebased to `4`

PR #318, opened 2020-09-04
DOCS Restructure markdown, remove wiki

chillu:
This is pretty much the only supported module that started with
storing its docs in the Github wiki at https://github.com/symbiote/silverstripe-queuedjobs/wiki.

Over time, this lead to fragmentation, and new documentation improvements
were just tucked on to the "quick overview" in the README,
as well as stray pages without much context in docs/en.
The wiki approach is inferior, because it requires edit permissions
and doesn't enable a pull request based workflow alongside other feature changes.

This PR is 80% copy paste, with a bit of rewrite mixed in (see individual commits).
I ported most of what was on the wiki, but there's some (near) duplication
which I've resolved.

After merging this PR, please disable the wiki functionality.


emteknetnz (approved):
Should this target 4.6 instead?

How do we disable a wiki (I didn't see an option, I probably lack privileges)

chillu:
> Should this target 4.6 instead?

I think targeting the default branch (`4`) is OK, it's the thing most people look at on github.com as the "documentation browser".

> How do we disable a wiki (I didn't see an option, I probably lack privileges)

https://docs.github.com/en/github/building-a-strong-community/disabling-wikis. If you can't see this, ping an admin? I don't have any rights on this repo.

emteknetnz:
@brynwhyman I don't have admin rights to remove the wiki, as shown in the instructions above.  Do you know who an admin would be? Would you be able to contact them asking them to remove the wiki?

PR #319, opened 2020-10-07
Update QueuedJobDescriptor.php

vinstah:
PHP Deprecated:  Array and string offset access syntax with curly braces


dhensby (approved):

muppsy007:
Hi there, is this likely to be merged into the 3.x tag? I'm getting the same Deprecation notice. I see there's a branch feature-php-7-4 that addresses it, but it's inaccessible to composer.

PR #320, opened 2020-11-02
BUG: Queued job service INIT and WAITING state lock release fix.

mfendeksilverstripe:
# Queued job service INIT and WAITING state lock release fix

This was split from https://github.com/symbiote/silverstripe-queuedjobs/pull/287

## Details

* added a `initialising_state_ttl` setting for `Initialising` to cover an edge case situation, this puts a hard limit on how long a job is allowed to stay in such state (it shouldn't stay in this state for long)
* job lock is released when job processing hits time execution limit or memory limit and goes to waiting state
* minor linting changes

## Failing unit tests

* This is an upstream error, not related to this PR

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -125,6 +125,15 @@ class QueuedJobService
     private static $worker_ttl = 'PT5M';
 
     /**
+     * Timeout value in seconds for the Initialising state
+     * if a job is stuck in this state longer than this value it's considered stalled
+     *
+     * @var int
+     * @config
+     */
+    private static $initialising_state_ttl = 120;
+
+    /**
      * Timestamp (in seconds) when the queue was started
      *
      * @var int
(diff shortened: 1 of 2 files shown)


nicole-ashley:
Thanks @mfendeksilverstripe!

src/Services/QueuedJobService.php
```diff
@@ -124,6 +124,15 @@ class QueuedJobService
...
     private static $worker_ttl = 'PT5M';
 
+    /**
+     * Timeout value in seconds for the Initialising state
+     * if a job is stuck in this state longer than this value it's considered stalled
+     *
+     * @var int
+     * @config
+     */
+    private static $initialising_state_ttl = 120;
```
Cheddam: Does it make sense to retain the same format across different TTL settings? See `$worker_ttl`.

```suggestion
    private static $initialising_state_ttl = 'PT2M';
```

Cheddam (commented):
This looks pretty much good to go - just one question around the configuration.

src/Services/QueuedJobService.php
mfendeksilverstripe: Good spotting, I'll update the configuration format 👍  thanks

mfendeksilverstripe:
@Cheddam Pushed up new changes (consistent configuration), please review.

PR #324, opened 2020-11-16
Fix build error from handling broken jobs

### LINKED ISSUE #316: Fatal error if exception thrown during job initialisation
> ERROR [Emergency]: Uncaught TypeError: Argument 2 passed to Symbiote\QueuedJobs\Services\QueuedJobService::handleBrokenJobException() must implement interface Symbiote\QueuedJobs\Services\QueuedJob, null given, called in /<siteroot>/vendor/symbiote/silverstripe-queuedjobs/src/Services/QueuedJobService.php on line 1038
IN GET dev/tasks/ProcessJobQueueTask

The `$job` variable is initialised to `null`: https://github.com/symbiote/silverstripe-queuedjobs/blob/4/src/Services/QueuedJobService.php#L798. If an exception is then thrown during `initialiseJob()`, then `$job` is still null, which triggers the fatal error when it’s then passed to `handleBrokenJobException`.

Simplest fix is just changing the method signature for `handleBrokenJobException` to make the second argument optional: `QueuedJob $job = null`

matt-in-a-hat:
Fixes #316 where an exception occurs when trying to instantiate the $job object, which means in this handler it expects a QueuedJob but gets null.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -984,7 +984,7 @@ class QueuedJobService
                 }
             } catch (\Throwable $e) {
                 // PHP 7 Error handling)
-                $this->handleBrokenJobException($jobDescriptor, $job, $e);
+                $this->handleBrokenJobException($jobDescriptor, $e, $job);
                 $broken = true;
             }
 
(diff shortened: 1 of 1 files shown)


emteknetnz:
Sorry can't merge this one, since it's a protected function changing a method signature would not be backwards compatible and would require a major release according to [semver](https://semver.org/) which we follow

Feel free to open an alternative pull-request that doesn't need a method signature change

PR #326, opened 2020-11-17
BUG: Exclude tasks run via a job from health check.

mfendeksilverstripe:
# Exclude tasks run via a job from health check

* Tasks which are run via job wrapper never have steps so the queue health check doesn't work properly
* This solution will exclude such tasks from the health check leaving the task processing run once without interruption
* If such task will timeout or is interrupted by other means the health check will not process such task

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/325

## Tests

Test are passing on my local. The broken tests seem to be an upstream error.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -28,6 +28,7 @@ use SilverStripe\Security\Security;
 use SilverStripe\Subsites\Model\Subsite;
 use Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor;
 use Symbiote\QueuedJobs\Interfaces\UserContextInterface;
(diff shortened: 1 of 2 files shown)


chillu:
Yep that looks great, thanks for looking at that so quickly Mojmir! I've since fixed it by increasing the `worker_ttl`, and just preparing a docs update to point out this footgun.

chillu:
Docs at https://github.com/symbiote/silverstripe-queuedjobs/pull/327

chillu:
The build failures should be fixed by https://github.com/symbiote/silverstripe-queuedjobs/pull/323. Happy with this PR, but can't merge myself (no permissions).

Cheddam (approved):
Thanks @mfendeksilverstripe!

PR #327, opened 2020-11-17
DOCS Call out step requirement to avoid staleness

chillu:
Context: https://github.com/symbiote/silverstripe-queuedjobs/issues/325

### DIFF
--- a/docs/en/troubleshooting.md
+++ b/docs/en/troubleshooting.md
@@ -40,7 +40,7 @@ A long running job _may_ fool the system into thinking it has gone away (ie the 
 `currentStep` hasn't been incremented). To avoid this scenario, you can set `$this->currentStep = -1` in your job's
 constructor, to prevent any health checks detecting the job.****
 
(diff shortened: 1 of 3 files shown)


docs/en/troubleshooting.md
```diff
@@ -54,13 +54,19 @@ Symbiote\QueuedJobs\Services\QueuedJobService:
   disable_health_check: true
 ```
 
-In addition to the config setting there is a task that can be used with a cron to ensure that unhealthy jobs are
-detected:
+Job health is checked automatically in queue processing.
+You might also need to disable the `CheckJobHealthTask` if it's set up as a cron job.
+
+Alternatively, you can increase the TTL before jobs a considered stalled:
```
Cheddam:
```suggestion
Alternatively, you can increase the TTL before jobs are considered stalled:
```

chillu:
@Cheddam Ready to merge, thanks for the review :)

Cheddam (approved):

PR #328, opened 2020-11-19
[4.6 BACKPORT] BUG: Exclude tasks run via a job from health check.

Cheddam:


emteknetnz:
There was some red bits in travis, though they only related to missing dev versions of silverstripe modules <=4.4

PR #329, opened 2020-11-20
Check Doorman log path is a valid path

andrewandante:
Just had a silent error here where it collapsed because it couldn't write to the path I set - so the Queue stalled at initialising.

### DIFF
--- a/src/Tasks/Engines/DoormanRunner.php
+++ b/src/Tasks/Engines/DoormanRunner.php
@@ -63,7 +63,7 @@ class DoormanRunner extends BaseRunner implements TaskRunnerEngine
             BASE_PATH . "/vendor/silverstripe/framework/cli-script.php dev/tasks/ProcessJobQueueChildTask"
         );
         $logPath = Environment::getEnv('SS_DOORMAN_LOGPATH');
-        if ($logPath && is_dir($logPath)) {
+        if ($logPath && is_dir($logPath) && is_writable($logPath)) {
             $manager->setLogPath($logPath);
         }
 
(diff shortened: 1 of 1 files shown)


andrewandante:
Possibly needs an `is_writeable()` check as well?

src/Tasks/Engines/DoormanRunner.php
```diff
@@ -63,7 +63,7 @@ public function runQueue($queue)
             BASE_PATH . "/vendor/silverstripe/framework/cli-script.php dev/tasks/ProcessJobQueueChildTask"
         );
         $logPath = Environment::getEnv('SS_DOORMAN_LOGPATH');
-        if ($logPath) {
+        if ($logPath && is_dir($logPath)) {
```
emteknetnz:
```suggestion
        if ($logPath && is_dir($logPath) && is_writable($logPath)) {
```

emteknetnz: What happens if the dir doesn't yet exist?  It's not automatically created somewhere via `mkdir()`?

emteknetnz (changes requested):
Left a suggestion to add is_writable() as you suggested

Also a question around mkdir

Happy to merge as is after those have been answered/actioned

src/Tasks/Engines/DoormanRunner.php
andrewandante: No it's not, it's just appended to the stdOut and stdErr calls if it's set.

emteknetnz (approved):
Travis failures are existing

PR #337, opened 2021-03-01
BUG: Allow zero values for integers and floats within queued job rule.

mfendeksilverstripe:
# Allow zero values for integers and floats within queued job rule

Zero values for integers and floats within queued job rule are needed to cover the cases where high memory or processor use should prevent any new jobs from being started. This is currently impossible to set even though it's internally supported within the `AsyncPHP\Doorman\Rules\InMemoryRules::canRunTask`.

This change-set allows zero values to be set for integers and floats.

## Related issues

https://github.com/symbiote/silverstripe-queuedjobs/issues/338

### DIFF
--- a/src/DataObjects/QueuedJobRule.php
+++ b/src/DataObjects/QueuedJobRule.php
@@ -48,7 +48,7 @@ class QueuedJobRule extends DataObject implements Rule
      */
     public function getProcesses()
     {
-        if ($this->getField('Processes')) {
+        if ($this->getField('Processes') !== null) {
             return $this->getField('Processes');
         }
 
@@ -74,7 +74,7 @@ class QueuedJobRule extends DataObject implements Rule
      */
     public function getMinimumProcessorUsage()
     {
-        if ($this->getField('MinimumProcessorUsage')) {
+        if ($this->getField('MinimumProcessorUsage') !== null) {
             return $this->getField('MinimumProcessorUsage');
         }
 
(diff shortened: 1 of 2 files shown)


maxime-rainville (commented):
Just clarifying the problem to make sure I understand. So you need to be able to set these values to `0`, but because all those getters return their default value when ever a "falsy" value is provided, your ZEROs are always ignored.

How would you actually use this in practice? You would create a Rule for a job and set the process to zero to block any other job from running while that job is running?

mfendeksilverstripe:
That's correct @maxime-rainville , here is some more information as requested:

Screen Shot 2021-03-09 at 7 26 21 AM

In the example above we have a rule saying that if your memory usage is between 75 and 100 you should create only one new process but what we actually want to do is to set it to 0 - if memory usage is between 75 and 100 we want to create 0 new processes. Same for processor use.

Screen Shot 2021-03-09 at 7 29 34 AM

Note that you can actually store a 0 on a `QueuedJobRule` model in the DB but when the value is fetched from the DB it's going through that getter and it will override the value.

Note that the [actual execution](https://github.com/asyncphp/doorman/blob/3.1.0/src/Rules/InMemoryRules.php#L59) of the `QueuedJobRule` is implemented in a way to check for nulls not falsy values.

maxime-rainville (changes requested):
That sound good. I presume that the screen shot you have in your last comment comes from a custom gridfield you implemented because I can't find it anywhere.

Can you add some unit test to cover that those getters handle `0` as you expect them to and return the correct value? 

Probably don't need to add a unit test to cover that providing a 0 ends up having the intended affects on the queues since that's handle by the third party package.

mfendeksilverstripe:
Yes, that's a custom `GridField`. Sure I'll add unit tests for that.

mfendeksilverstripe:
@maxime-rainville Added the unit tests as requested. Please review.

maxime-rainville (approved):
Looks good

PR #342, opened 2021-03-26
[BUGFIX #317] Make queueJob threadsafe

### LINKED ISSUE #317: Job signature system is not thread safe
# Job signature system is not thread safe

Running multiple instances of dev tasks which create the same jobs may result in multiple copies of the same job instance with the same signature.

Running multiple dev tasks at the same time is not intentional but this may happen in some situations. For example:

* running a dev task with long execution time via browser may result in the browser thinking that the page is unresponsive and firing a redundant request
* running a dev task with long execution time as a job may result in the queue runner thinking the task job is stale and resuming it

## Suggested solution

Inside `QueuedJobService::queueJob()` wrap the job creation inside transaction (use `withTransaction`) and change the check for existing job to use the `SELECT ... FOR UPDATE` approach (similar to job locking).

## PR
- [ ] https://github.com/symbiote/silverstripe-queuedjobs/pull/342

danaenz:
Adds a transaction lock to JobDescriptor lookup to prevent job creation with duplicate signatures.

## Functionality improvements
 
- Wrapped queueJob read/write functionality in withTransaction()
- Added exception handling

## Readability

- Separated functionality into protected methods for readability

Fixes [Issue 317](https://github.com/symbiote/silverstripe-queuedjobs/issues/317)

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -235,64 +235,30 @@ class QueuedJobService
     {
         $signature = $job->getSignature();
 
-        // see if we already have this job in a queue
-        $filter = [
-            'Signature' => $signature,
-            'JobStatus' => [
-                QueuedJob::STATUS_NEW,
-                QueuedJob::STATUS_INIT,
-            ],
-        ];
+        // Create the initial object
+        $jobDescriptor = $this->createJobDescriptor($job, $signature, $startAfter, $userId, $queueName);
 
-        $existing = QueuedJobDescriptor::get()
-            ->filter($filter)
-            ->first();
-
-        if ($existing && $existing->ID) {
-            return $existing->ID;
+        try {
+            return $this->findOrMakeJobDescriptorFromSignature($signature, $job, $jobDescriptor, $startAfter);
+        } catch (\Throwable $e) {
+            // note that error here may not be an issue as failing to acquire a job lock is a valid state
+            // which happens when other process claimed the job lock first
+            $this->getLogger()->debug(
+                sprintf(
+                    '[%s] - Queued Jobs - Failed to acquire job lock %s %d %s',
+                    DBDatetime::now()->Rfc2822(),
+                    $e->getMessage(),
+                    $signature,
+                    PHP_EOL
+                ),
+                [
+                    'file' => __FILE__,
+                    'line' => __LINE__,
+                ]
+            );
         }
 
-        $jobDescriptor = new QueuedJobDescriptor();
-        $jobDescriptor->JobTitle = $job->getTitle();
-        $jobDescriptor->JobType = $queueName ? $queueName : $job->getJobType();
-        $jobDescriptor->Signature = $signature;
-        $jobDescriptor->Implementation = get_class($job);
-        $jobDescriptor->StartAfter = $startAfter;
-
-        // no user provided - fallback to job user default
-        if ($userId === null && $job instanceof UserContextInterface) {
-            $userId = $job->getRunAsMemberID();
-        }
-
-        // still no user - fallback to current user
-        if ($userId === null) {
-            if (Security::getCurrentUser() && Security::getCurrentUser()->exists()) {
-                // current user available
-                $runAsID = Security::getCurrentUser()->ID;
-            } else {
-                // current user unavailable
-                $runAsID = 0;
-            }
-        } else {
-            $runAsID = $userId;
-        }
-
-        $jobDescriptor->RunAsID = $runAsID;
-
-        // use this to populate custom data columns before job is queued
-        // note: you can pass arbitrary data to your job and then move it to job descriptor
-        // this is useful if you need some data that needs to be exposed as a separate
-        // DB column as opposed to serialised data
-        $this->extend('updateJobDescriptorBeforeQueued', $jobDescriptor, $job);
-
-        // copy data
-        $this->copyJobToDescriptor($job, $jobDescriptor);
-
-        $jobDescriptor->write();
-
-        $this->startJob($jobDescriptor, $startAfter);
-
-        return $jobDescriptor->ID;
+        return false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Services/QueuedJobService.php
```diff
@@ -235,64 +235,30 @@ public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $qu
...
-                // current user unavailable
-                $runAsID = 0;
-            }
-        } else {
-            $runAsID = $userId;
+        try {
+            return $this->findOrMakeJobDescriptorFromSignature($signature, $job, $jobDescriptor, $startAfter);
+        } catch (\Throwable $e) {
+            // note that error here may not be an issue as failing to acquire a job lock is a valid state
+            // which happens when other process claimed the job lock first
```
mfendeksilverstripe, line 245: I think the comments should be updated to describe the job signature, not the job lock as that is a different feature. Please also update the log message so this feature is not confused with the job locking.

brynwhyman:
Hi @danaenz a quick follow up to see if you're still planning on addressing the feedback in this PR? :)

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+     * Using a job signature, returns the JobDescriptor ID and whether the
+     * job descriptor is new or existing
+     *
+     * @param string $signature
+     * @param QueuedJob $job
+     * @param QueuedJobDescriptor $jobDescriptor
+     * @param null|string $startAfter
+     * @return int|null
+     */
+    protected function findOrMakeJobDescriptorFromSignature($signature, $job, $jobDescriptor, $startAfter)
```
emteknetnz, line 313: This should be made private to reduce the API surface.  There's an extension hook available if it needs to be extended.

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+
+    /**
+     * @param QueuedJob $job
+     * @param string $signature
+     * @param null $startAfter
+     * @param null $userId
+     * @param null $queueName
+     * @return QueuedJobDescriptor
+     */
+    protected function createJobDescriptor(
```
emteknetnz, line 357: Should change this from protected to private

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+     * @param QueuedJob $job
+     * @param QueuedJobDescriptor $jobDescriptor
+     * @param null|string $startAfter
+     * @return int|null
+     */
+    protected function findOrMakeJobDescriptorFromSignature($signature, $job, $jobDescriptor, $startAfter)
+    {
+        // Start a transaction which will hold until we have a lock on this signature.
+        return DB::get_conn()->withTransaction(function () use ($signature, $job, $jobDescriptor, $startAfter) {
+            $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "Signature" = ? FOR UPDATE';
```
emteknetnz, line 317: Bit of an edge case, though should read the tableName of QueuedJobDescriptor first incase it's been changed (table names are configuraable

`$tableName = Config::inst()->get(QueuedJobDescriptor::class, 'table_name', Config::UNINHERITED);`

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+        $jobDescriptor->StartAfter = $startAfter;
+
+        // no user provided - fallback to job user default
+        if ($userId === null && $job instanceof UserContextInterface) {
+            $userId = $job->getRunAsMemberID();
+        }
+
+        // still no user - fallback to current user
+        if ($userId === null) {
+            if (Security::getCurrentUser() && Security::getCurrentUser()->exists()) {
```
emteknetnz, line 379: Should put this to a variable so that it doesn't need to be called 3 separate times times
`$member = Security::getCurrentUser()`

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+     * @param null $startAfter
+     * @param null $userId
+     * @param null $queueName
+     * @return QueuedJobDescriptor
+     */
+    protected function createJobDescriptor(
+        QueuedJob $job,
+        $signature,
+        $startAfter = null,
+        $userId = null,
```
emteknetnz, line 361:
```suggestion
        $userID = null,
```
Convention is to capitilise the d in ID

Also update the references in this function to `$userID`

src/Services/QueuedJobService.php
```diff
@@ -235,64 +235,30 @@ public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $qu
...
-
-        // copy data
-        $this->copyJobToDescriptor($job, $jobDescriptor);
-
-        $jobDescriptor->write();
-
-        $this->startJob($jobDescriptor, $startAfter);
-
-        return $jobDescriptor->ID;
+        return false;
```
emteknetnz, line 261:
```suggestion
        return 0;
```
PHPDOC function signature is to return an int

src/Services/QueuedJobService.php
```diff
@@ -235,64 +235,30 @@ public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $qu
...
-        $jobDescriptor->Signature = $signature;
-        $jobDescriptor->Implementation = get_class($job);
-        $jobDescriptor->StartAfter = $startAfter;
-
-        // no user provided - fallback to job user default
-        if ($userId === null && $job instanceof UserContextInterface) {
-            $userId = $job->getRunAsMemberID();
-        }
+        // Create the initial object
+        $jobDescriptor = $this->createJobDescriptor($job, $signature, $startAfter, $userId, $queueName);
```
emteknetnz, line 239: This is kind of weird creating the job and then passing it to `findOrMakeJobDescriptorFromSignature`

Seems like we should call `createJobDescriptor()` inside `findOrMakeJobDescriptorFromSignature` on if we were unable to find a jobDescriptor from the signature

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+                // note: you can pass arbitrary data to your job and then move it to job descriptor
+                // this is useful if you need some data that needs to be exposed as a separate
+                // DB column as opposed to serialised data
+                $this->extend('updateJobDescriptorBeforeQueued', $jobDescriptor, $job);
+
+                // copy data
+                $this->copyJobToDescriptor($job, $jobDescriptor);
+
+                // Write the record
+                $jobDescriptorID = $jobDescriptor->write();
```
emteknetnz, line 340: `$jobDescriptorID` variable is unused?

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+
+                $this->startJob($jobDescriptor, $startAfter);
+            } else {
+                $jobDescriptorID = $ID;
```
emteknetnz, line 344: `$jobDescriptorID` appears to be unused, so we can just omit this?

src/Services/QueuedJobService.php
```diff
@@ -334,6 +300,98 @@ public function isAtMaxJobs()
...
+     * @param string $signature
+     * @param QueuedJob $job
+     * @param QueuedJobDescriptor $jobDescriptor
+     * @param null|string $startAfter
+     * @return int|null
+     */
+    protected function findOrMakeJobDescriptorFromSignature($signature, $job, $jobDescriptor, $startAfter)
+    {
+        // Start a transaction which will hold until we have a lock on this signature.
+        return DB::get_conn()->withTransaction(function () use ($signature, $job, $jobDescriptor, $startAfter) {
```
emteknetnz, line 316: Do we need the return? `Database::withTransaction()` doesn't have a return value?

emteknetnz (changes requested):
Approach looks good overall

mfendeksilverstripe:
@danaenz Please review the feedback on this PR and let me know if you need help with it ;-) Would be keen on getting this over the line.

PR #344, opened 2021-04-13
[BUGFIX] Parse build task class correctly

coldblooded01:


emteknetnz (changes requested):
Could you please provide a description of what you're trying to fix here

michalkleiner:
Hi @emteknetnz, the OP doesn't have access to the repo anymore. The issue we had was that the check was failing for some task classes. Looking at the suggested fix, it was either there were some white characters around the actual class name or it was checking against itself when the BuildTask class was injected over.

The suggested change is quite safe from my perspective. I could potentially dig up the exact reason but not sure how beneficial that would be anyway.

emteknetnz:
I don't think `is_a` is better than `is_subclass_of` here, because `BuildTask` is an abstract class.  This shouldn't be getting injected over injected over.

Seems like all this needs is a trim() added? ... though I cannot think of why there would ever be additional space characters around `$this->TaskClass`

michalkleiner:
You're probably right. We ended deploying the fork with this change on a single site so I'll look into why it needed this fix and if the trim was enough (and/or if it should be fixed somewhere else).

PR #350, opened 2021-07-07
Remove stalled job if implementation not found

wilr:
If a job class is renamed or removed, remove the related stalled jobs rather than throwing an exception (which prevents the rest of the jobs from running).

### DIFF
--- a/src/Jobs/DoormanQueuedJobTask.php
+++ b/src/Jobs/DoormanQueuedJobTask.php
@@ -18,7 +18,7 @@ class DoormanQueuedJobTask implements Task, Expires, Process, Cancellable
     protected $id;
 
     /**
-     * @var QueuedJobDescriptor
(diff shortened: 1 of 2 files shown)


dhensby:
linting issues :)

wilr:
@dhensby lint issue fixed

src/Jobs/DoormanQueuedJobTask.php
```diff
@@ -59,7 +59,7 @@ public function setId($id)
     /**
      * @return QueuedJobDescriptor
      */
-    public function getDescriptor()
+    public function getDescriptor(): ?QueuedJobDescriptor
```
emteknetnz: Technically an API change, could you remove this change

emteknetnz (changes requested):

emteknetnz (approved):
Good

PR #354, opened 2021-08-18
releaseJobLock when marking a job as broken

### LINKED ISSUE #339: Broken jobs won’t resume
When a job (more than 1 “step”, not sure if that’s relevant) throws an exception, it’s marked with a JobStatus of `Broken`. The admin UI shows a button to resume the job:

Screenshot 2021-03-05 at 09 06 30

Clicking this changes the JobStatus back to `Waiting`, but the job never tries to run again. The only way I can get it to run is by clicking to edit the job and deleting the hash(?) in the “Worker” field.

andrewandante:
Fixes #339 

Also added a few visibility things on some constants.

### DIFF
--- a/src/Services/QueuedJob.php
+++ b/src/Services/QueuedJob.php
@@ -14,32 +14,32 @@ interface QueuedJob
      * Always run immediate jobs as soon as possible
      * @var string
      */
-    const IMMEDIATE = '1';
+    public const IMMEDIATE = '1';
 
     /**
      * Queued jobs may have some processing to do, but should be pretty quick
      * @var string
      */
-    const QUEUED = '2';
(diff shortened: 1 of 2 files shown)


src/Services/QueuedJob.php
```diff
@@ -14,32 +14,32 @@ interface QueuedJob
      * Always run immediate jobs as soon as possible
      * @var string
      */
-    const IMMEDIATE = '1';
+    public const IMMEDIATE = '1';
```
emteknetnz, line 17: default visibility in public, so this isn't an API change

emteknetnz (approved):
Looks good to me

PR #365, opened 2022-02-02
NEW: Default jobs / queue type configuration.

mfendeksilverstripe:
# NEW: Default jobs / queue type configuration.

Added a new configuration option to specify queue type for default jobs. This is handy in case multiple queues are used and you want only one of them to run default jobs.

### DIFF
--- a/docs/en/default-jobs.md
+++ b/docs/en/default-jobs.md
@@ -75,7 +75,7 @@ SilverStripe\Core\Injector\Injector:
           # Set the email address to send the alert to if not set site admin email is used OPTIONAL
           email: 'admin@example.com'
           # Make this job specific to only certain queue type (see QueuedJob interface)
-          queue: 1 # This would make this job be picked up only by Immediate queue
+          jobType: 2 # This would make this job be picked up only by the QUEUED queue
         # Minimal implementation will send alerts but not recreate
         AnotherTitle:
           type: 'AJob'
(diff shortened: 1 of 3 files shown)


emteknetnz (changes requested):
Looks good

Could you please add QueuedJobServiceTest.php and add a quick test for this.  Update the config and then assert the `->defaultJobs` array has what's expected

mfendeksilverstripe:
@emteknetnz I've added a test for this, please review.

docs/en/default-jobs.md
```diff
@@ -74,6 +74,8 @@ SilverStripe\Core\Injector\Injector:
           recreate: 1
           # Set the email address to send the alert to if not set site admin email is used OPTIONAL
           email: 'admin@example.com'
+          # Make this job specific to only certain queue type (see QueuedJob interface)
+          queue: 1 # This would make this job be picked up only by Immediate queue
```
emteknetnz:
```suggestion
          jobType: 2 # This would make this job be picked up only by the QUEUED queue
```

Rather than use `queue` which matches the variable `$queue`, I think we should using `jobType` key as it matches the interface

**QueuedJob.php**
```php
    /**
     * What type of job is this? Options are
     * - QueuedJob::IMMEDIATE
     * - QueuedJob::QUEUED
     * - QueuedJob::LARGE
     */
    public function getJobType();
```

Also I think we should change the value in the readme from `1` to `2` (and the corresponding comment) because that's the queue that gets run on `vendor/bin/sake dev/tasks/ProcessJobQueueTask`, so it's much easier for user to validate that this is working.

emteknetnz (changes requested):
Have tested locally, it works.  I think we should update the key name used as well as the jobType/queue in the readme example

mfendeksilverstripe:
@emteknetnz I've updated the name as requested, please review.

emteknetnz (approved):
Tested locally, works as it should

PR #374, opened 2022-05-31
FIX: Correctly log errors during job initialisation (fixes #316)

### LINKED ISSUE #316: Fatal error if exception thrown during job initialisation

lozcalver:
The alternative to this is changing the the signature of `handleBrokenJobException()` to make the second param optional with `= null`, which is technically a BC-break. This method of instantiating a job object is exactly what `initialiseJob` does:

https://github.com/symbiote/silverstripe-queuedjobs/blob/84fc3e2bc1c0f2fc3b6df9851cdb16a4c4606bc1/src/Services/QueuedJobService.php#L676-L683

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -31,6 +31,7 @@ use Symbiote\QueuedJobs\Interfaces\UserContextInterface;
 use Symbiote\QueuedJobs\Jobs\RunBuildTaskJob;
 use Symbiote\QueuedJobs\QJUtils;
 use Symbiote\QueuedJobs\Tasks\Engines\TaskRunnerEngine;
+use Throwable;
 
 /**
  * A service that can be used for starting, stopping and listing queued jobs.
@@ -753,7 +754,7 @@ class QueuedJobService
             });
 
             return true;
-        } catch (\Throwable $e) {
+        } catch (Throwable $e) {
             // note that error here may not be an issue as failing to acquire a job lock is a valid state
             // which happens when other process claimed the job lock first
             $this->getLogger()->debug(
@@ -829,7 +830,14 @@ class QueuedJobService
             $job = null;
 
             try {
-                $job = $this->initialiseJob($jobDescriptor);
+                try {
+                    $job = $this->initialiseJob($jobDescriptor);
+                } catch (Throwable $e) {
+                    $this->handleJobInitialisationException($jobDescriptor, $e);
+                    $broken = true;
+                    $this->finaliseLogging($logger);
+                    return;
+                }
 
                 // get the job ready to begin.
                 if (!$jobDescriptor->JobStarted) {
@@ -911,7 +919,7 @@ class QueuedJobService
 
                         try {
                             $job->process();
-                        } catch (\Throwable $e) {
+                        } catch (Throwable $e) {
                             $logger->error($e->getMessage(), ['exception' => $e]);
                             $this->markJobAsBroken($jobDescriptor);
                             $this->extend('updateJobDescriptorAndJobOnException', $jobDescriptor, $job, $e);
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Are there some steps I can follow to ensure a job breaks during initialisation in order to test this?
And is this something we could write unit tests for?

> This method of instantiating a job object is exactly what `initialiseJob` does

Maybe that should be extracted into its own method so we can reuse the same code? That way if the way we instantiate these changes, we can remain consistent.

lozcalver:
To recreate this you just need a job to throw an exception in its `setup()` method.

I don’t think this fix is complete after all because if the exception shown above (`Implementation $impl no longer exists`) is thrown then the exception still won’t be logged properly. I wonder if a better approach would be to wrap this in its own try/catch:

https://github.com/symbiote/silverstripe-queuedjobs/blob/c370b278f7a417bbc002a00891a897aec3f71747/src/Services/QueuedJobService.php#L832

That would allow us to handle the exception-during-initialisation flow differently, so no need to adjust method signatures and break BC

lozcalver:
Okay I’ve pushed an alternative solution

src/Services/QueuedJobService.php
```diff
@@ -829,7 +830,13 @@ public function runJob($jobId)
             $job = null;
 
             try {
-                $job = $this->initialiseJob($jobDescriptor);
+                try {
+                    $job = $this->initialiseJob($jobDescriptor);
+                } catch (Throwable $e) {
+                    $this->handleJobInitialisationException($jobDescriptor, $e);
+                    $broken = true;
+                    goto finalise_logging;
```
ChloeSartorelli: I think it would probably be better if the code after `finalise_logging` became its own method which we can call here instead.

src/Services/QueuedJobService.php
```diff
@@ -1044,19 +1056,27 @@ protected function withNestedState(callable $callback)
         }
     }
 
+    protected function handleJobInitialisationException(QueuedJobDescriptor $jobDescriptor, Throwable $e)
```
ChloeSartorelli, line 1059: Why is this separate from `handleBrokenJobException()`?

lozcalver: Because we don’t have a `$job` instance to pass to `handleBrokenJobException()`, and changing the method signature to make that argument optional is a BC break (plus it’d mean putting an optional argument ahead of a required argument, which is a bit yucky 😛)

ChloeSartorelli: Couldn't you just pass in `null`?

lozcalver: Nope, the argument is typehinted as `QueuedJob $job` so passing `null` will error

ChloeSartorelli: Ahh yup fair enough.

GuySartorelli:
Looks good from a code perspective. Will try to make some time tomorrow to test it locally just to be sure it does catch those errors.

GuySartorelli (commented):
Gotta love when you see a comment from yourself a week later that says "tomorrow" on it.... sorry about that.

Looks good locally except:

## Doesn't log the exception to the job's messages
This does catch and mark the exception and marks the job as broken which is a big improvement on the current state of things - but it doesn't add the exception information to the job's messages like exceptions later in the lifecycle do.

Looks like that's because `addJobHandlersToLogger()` isn't called until later, which expects a `QueuedJob` instance (which we obviously don't have at this stage) to pass to `QueuedJobHandler`.

I don't think `QueuedJobHandler` should explicitly require a `QueuedJob` instance (messages are stored on the `QueuedJobDescriptor` anyway which we _do_ have) - but since it does, we can't just add that handler to the logger.

## Next steps
I think we should create a new handler (to either be replaced by `QueuedJobHandler` or else to be used for this specific situation - preferable the former with `QueuedJobHandler` being deprecated) which _doesn't_ require a `QueuedJob` instance.

I _don't_ know if that should be done as part of this PR (it could be considered part of "correctly" logging errors) or as a follow-up. If it's a follow-up, then I'm happy with this PR as-is. It is at least an improvement on the current state of things.

GuySartorelli:
@kinglozzer do you have any thoughts about whether logging the message should be done with or after thIs PR?

lozcalver:
I think it’d need to be done separately - we can’t switch out `QueuedJobHandler` for one that doesn’t require a `QueuedJob` instance without breaking BC (e.g. if someone is using Injector to inject their own handler that uses the `QueuedJob` object)

GuySartorelli (approved):
Works well locally. Doesn't log these errors in the message for the job, but that can be follow-up work as discussed.
At least it logs them correctly _in the logger_ which is an improvement over the existing behaviour.

GuySartorelli:
I have created https://github.com/symbiote/silverstripe-queuedjobs/issues/384 for the follow-up work of passing the error as a message into the job to be viewed in the CMS UI.

GuySartorelli:
Tagged as [4.10.1](https://github.com/symbiote/silverstripe-queuedjobs/releases/tag/4.10.1)

PR #383, opened 2022-08-19
feat: Allow ProcessManager PHP binary to be configurable via environment variable

LeviWB:
…ent variable using Doorman.


GuySartorelli (approved):
Seems like a good change, works locally. Thanks!
I don't think this can be meaningfully tested with unit tests so approving as-is.

GuySartorelli:
Merge on green.

PR #385, opened 2022-09-08
ENH Add a Type and Status filter to Jobs

maxime-rainville:
By default, you can only filter queued jobs by title, which is kind of stupid.

This adds Type and Status dropdown filters.

image

### DIFF
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -132,14 +132,6 @@ class QueuedJobDescriptor extends DataObject
     ];
 
     /**
-     * @var array
-     */
-    private static $field_labels = [
-        'JobStatus' => 'Status',
-        'JobType' => 'Queue Type'
-    ];
-
-    /**
      * @var string
      */
     private static $default_sort = 'Created DESC';
@@ -590,7 +582,7 @@ class QueuedJobDescriptor extends DataObject
         $statuses = $this->getJobStatusValues();
         return DropdownField::create(
             'JobStatus',
-            $this->fieldLabel('JobStatus'),
+            _t(__CLASS__ . '.TABLE_STATUS', 'Status'),
             array_combine($statuses ?? [], $statuses ?? [])
         );
     }
@@ -602,7 +594,7 @@ class QueuedJobDescriptor extends DataObject
     {
         return DropdownField::create(
             'JobType',
-            $this->fieldLabel('JobType'),
+            _t(__CLASS__ . '.JOB_TYPE', 'Job Type'),
             $this->getJobTypeValues()
         );
     }
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -19,6 +19,7 @@ use SilverStripe\Forms\TextField;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use SilverStripe\ORM\FieldType\DBField;
+use SilverStripe\ORM\Filters\ExactMatchFilter;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use Symbiote\QueuedJobs\Services\QueuedJob;
@@ -126,6 +127,16 @@ class QueuedJobDescriptor extends DataObject
      */
     private static $searchable_fields = [
         'JobTitle',
+        'JobStatus' => ExactMatchFilter::class,
+        'JobType' => ExactMatchFilter::class
+    ];
+
+    /**
+     * @var array
+     */
+    private static $field_labels = [
+        'JobStatus' => 'Status',
+        'JobType' => 'Queue Type'
     ];
 
     /**
@@ -393,9 +404,7 @@ class QueuedJobDescriptor extends DataObject
     public function getCMSFields()
     {
         $fields = parent::getCMSFields();
-        $statuses = $this->getJobStatusValues();
         $runAs = $fields->fieldByName('Root.Main.RunAsID');
-
         $fields->removeByName([
             'Expiry',
             'Implementation',
@@ -432,8 +441,8 @@ class QueuedJobDescriptor extends DataObject
                 )
             ),
             $jobTitle = TextField::create('JobTitle', 'Title'),
-            $status = DropdownField::create('JobStatus', 'Status', array_combine($statuses ?? [], $statuses ?? [])),
-            $jobType = DropdownField::create('JobType', 'Queue type', $this->getJobTypeValues()),
+            $status = $this->buildJobStatusField(),
+            $jobType = $this->buildJobTypeField(),
             $runAs,
             $startAfter = DatetimeField::create('StartAfter', 'Scheduled Start Time'),
             HeaderField::create('JobTimelineTitle', 'Timeline'),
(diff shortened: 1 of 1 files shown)


src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -126,6 +127,8 @@ class QueuedJobDescriptor extends DataObject
      */
     private static $searchable_fields = [
```
maxime-rainville, line 128: Alternatively, we could remove $searchable_fields altogether and rely on the the default. However, you end up with all the summary fields in there. Most of them are date field which won't provide value unless you implement a range filter.

In any case, Status and Type seem like the most useful filter anyway.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -432,8 +433,8 @@ public function getCMSFields()
                 )
             ),
             $jobTitle = TextField::create('JobTitle', 'Title'),
-            $status = DropdownField::create('JobStatus', 'Status', array_combine($statuses ?? [], $statuses ?? [])),
-            $jobType = DropdownField::create('JobType', 'Queue type', $this->getJobTypeValues()),
+            $status = $this->buildJobStatusField(),
+            $jobType = $this->buildJobTypeField(),
```
maxime-rainville, line 437: Rather than duplicate the Dropdown creation logic below, I figured it made more sense to create private functions to keep everything aligned.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -585,4 +611,17 @@ private function getWorkerExpiry()
 
         return $time->getTimestamp() - $now->getTimestamp();
     }
+
+    public function getDefaultSearchContext()
```
maxime-rainville: Because JobStatus and JobType are VARCHARs rather than ENUM, they will default to Text field instead of dropdowns.

This bit of logic force them to use drop down.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -573,6 +574,31 @@ public function getCMSFields()
...
+        $statuses = $this->getJobStatusValues();
+        return DropdownField::create(
+            'JobStatus',
+            'Status',
+            array_combine($statuses ?? [], $statuses ?? [])
+        );
+    }
+
+    /**
+     * Generate a drop down field wit ha list of possible job types
```
ChloeSartorelli:
```suggestion
     * Generate a drop down field with a list of possible job types
```

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -585,4 +611,17 @@ private function getWorkerExpiry()
 
         return $time->getTimestamp() - $now->getTimestamp();
     }
+
+    public function getDefaultSearchContext()
+    {
+        $context =  parent::getDefaultSearchContext();
+
+        $context->addField($this->buildJobStatusField()->setEmptyString(''));
```
ChloeSartorelli: Fields should be added in `scaffoldSearchFields()`

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -585,4 +611,17 @@ private function getWorkerExpiry()
...
+        $context->addFilter(ExactMatchFilter::create('JobStatus'));
```
ChloeSartorelli: You can define filters in the searchable_fields array, which would probably be preferable here.
https://docs.silverstripe.org/en/4/developer_guides/model/scaffolding/#specify-a-form-field-or-search-filter

GuySartorelli (changes requested):

src/DataObjects/QueuedJobDescriptor.php
maxime-rainville: Done

src/DataObjects/QueuedJobDescriptor.php
maxime-rainville: Done

src/DataObjects/QueuedJobDescriptor.php
maxime-rainville: Done

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -585,4 +619,12 @@ private function getWorkerExpiry()
 
         return $time->getTimestamp() - $now->getTimestamp();
     }
+
+    public function scaffoldSearchFields($_params = null)
+    {
+        $fields = parent::scaffoldSearchFields($_params);
+        $fields->push($this->buildJobStatusField()->setEmptyString(''));
+        $fields->push($this->buildJobTypeField()->setEmptyString(''));
```
ChloeSartorelli, line 619: Does this not result in duplicated fields? I'd have thought fields for these columns would already be scaffolded (so `$fields->replaceField()` would be appropriate)?

maxime-rainville: FieldList doesn't let you have two fields with the same name. So pushing a field with a name that is already present in the fieldlist remove the pre-existing one. I think `replaceField` implicitly allows you to rename the field.

In this context, `push` seems cleaner to me but it's more of an aesthetic preference.

ChloeSartorelli: > I think replaceField implicitly allows you to rename the field.

It literally replaces the field outright. Yes, that does mean to can rename the field if you want to, but that's irrelevant to what's happening here.

`replaceField()` seems cleaner to be because that's what you're _actually_ doing - you're replacing the field which has already been scaffolded. But like you say since `push()` implicitly removes the existing field, it's a subjective aesthetic choice.

I'm not fussed either way - I leave the decision to you and to whoever ends up doing a proper review of this after the card has been brought into a sprint.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -573,6 +582,31 @@ public function getCMSFields()
...
+    }
+
+    /**
+     * Generate a drop down field with a list of possible job types
+     */
+    private function buildJobTypeField(): DropdownField
+    {
+        return DropdownField::create(
+            'JobType',
+            $this->fieldLabel('JobType'),
```
ChloeSartorelli: The label for this in the summary fields is "Job Type" and it uses `_t(__CLASS__ . '.JOB_TYPE', 'Job Type')` instead of field label - we should probably use that here as well so it's clear they are the same field.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -573,6 +582,31 @@ public function getCMSFields()
...
 
+    /**
+     * Generate a Dropdown field with the list of possible job
+     */
+    private function buildJobStatusField(): DropdownField
+    {
+        $statuses = $this->getJobStatusValues();
+        return DropdownField::create(
+            'JobStatus',
+            $this->fieldLabel('JobStatus'),
```
ChloeSartorelli:
```suggestion
            _t(__CLASS__ . '.TABLE_STATUS', 'Status'),
```
This is what's used in `summaryFields()` so it will keep localisations consistent.

GuySartorelli (changes requested):

GuySartorelli:
Works fine locally, just a couple small changes to make.

src/DataObjects/QueuedJobDescriptor.php
maxime-rainville: I'll leave it as-is.

maxime-rainville:
I'm now using the proper translation keys

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -126,6 +127,16 @@ class QueuedJobDescriptor extends DataObject
...
+        'JobStatus' => ExactMatchFilter::class,
+        'JobType' => ExactMatchFilter::class
+    ];
+
+    /**
+     * @var array
+     */
+    private static $field_labels = [
+        'JobStatus' => 'Status',
+        'JobType' => 'Queue Type'
```
ChloeSartorelli: This isn't needed now that we're using the translation strings.

GuySartorelli (changes requested):

GuySartorelli (approved):
Merge on green

PR #390, opened 2022-11-08
(DOCS) Added quotes around value in yml

lars-lemon8:
In more recent versions of the framework , the queueRunner value needs to be enclosed in quotes.

Otherwise:

`Fatal error: Uncaught Symfony\Component\Yaml\Exception\ParseException: The reserved indicator "%" cannot start a plain scalar; you need to quote the scalar at line 4 (near "queueRunner: %$DoormanRunner").`

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ After: '#queuedjobsettings'
 SilverStripe\Core\Injector\Injector:
   Symbiote\QueuedJobs\Services\QueuedJobService:
     properties:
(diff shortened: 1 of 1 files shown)


dhensby (approved):

michalkleiner (approved):

GuySartorelli (approved):
Merging before CI finishes because the readme doesn't affect that anyway.

PR #391, opened 2022-11-14
FIX grabMutex job lock query on SQLite

dmipeck:
Running a job using an SQLite database causes the update query at `src/Services/QueuedJobService.php:747` to throw an invalid query error. Using the ORM instead of raw database queries prevents this.

### DIFF
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -729,7 +729,8 @@ class QueuedJobService
         try {
             // Start a transaction which will hold until we have a lock on this descriptor.
             DB::get_conn()->withTransaction(function () use ($descriptorId) {
-                $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "ID" = %s AND "Worker" IS NULL FOR UPDATE';
+                $forUpdate = DB::getConfig()['type'] == 'SQLite3Database' ? '' : ' FOR UPDATE';
+                $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "ID" = %s AND "Worker" IS NULL' . $forUpdate;
 
                 $row = DB::query(sprintf($query, Convert::raw2sql($descriptorId)))->first();
 


src/Services/QueuedJobService.php
```diff
@@ -729,33 +729,27 @@ protected function grabMutex(QueuedJobDescriptor $jobDescriptor)
         try {
             // Start a transaction which will hold until we have a lock on this descriptor.
             DB::get_conn()->withTransaction(function () use ($descriptorId) {
-                $query = 'SELECT "ID" FROM "QueuedJobDescriptor" WHERE "ID" = %s AND "Worker" IS NULL FOR UPDATE';
+                $jobLockList = QueuedJobDescriptor::get()->filter([
```
ChloeSartorelli: The "FOR UPDATE" part of this query is vitally important (see https://www.cockroachlabs.com/blog/select-for-update/)

The ORM currently doesn't have functionality to handle this additional syntax. Any solution will need to either implement that into the ORM or otherwise produce SQL that includes this funtionality and is valid for sqlite.

GuySartorelli (changes requested):

src/Services/QueuedJobService.php
michalkleiner: The linked article says....
> For example, [MySQL supports SELECT FOR UPDATE](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html), but [SQLite does not](https://sqlite.org/lang_select.html).

ChloeSartorelli: I didn't read that far through to be honest... in that case I think we just simply can't support this piece of functionality with sqlite...

I'm not sure whether in that case we want to have some sort of feature detection to not try using locks when sqlite is used, or if we should just add something to the readme saying it's not usable with sqlite. But either way I think this PR is not the way forward.

dmipeck: > SQLite does not need SELECT FOR UPDATE because initiating a transaction locks the entire database

It seems like it won't be an issue to just drop the "FOR UPDATE" part of the query for SQLite?

ChloeSartorelli: Explicitly and only for SQLite (or other db connections that don't need it), yes. This PR drops it for every db.

I think we either need an ORM abstraction for it (where for anything that supports locking it adds the locking part to the query, and for anything that doesn't need it it just does nothing) OR we need some explicit feature detection in here.

I'd be more in favour of the former, though that will obviously require work in both framework and the sqlite module.

michalkleiner:
@dmipeck can you rebase your branch as it contains unrelated changes, please? Thanks

GuySartorelli:
I don't have time to set up sqlite in my dev environment - if someone can show me that this works I'll be okay with it, I think.

GuySartorelli (approved):
LGTM.
Retargetted to `4.12` so it can be released as a patch.

GuySartorelli:
This will be automatically tagged by GitHub actions

PR #395, opened 2023-01-19
FIX Handle array style email config

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10618

### DIFF
--- a/src/Services/EmailService.php
+++ b/src/Services/EmailService.php
@@ -2,6 +2,7 @@
 
 namespace Symbiote\QueuedJobs\Services;
 
+use InvalidArgumentException;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\Email\Email;
 use SilverStripe\Core\Config\Config;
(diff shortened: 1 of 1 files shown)


src/Services/EmailService.php
```diff
@@ -90,13 +91,27 @@ public function createStalledJobReport(string $subject, string $message, int $jo
...
+    /**
+     * Handle ['admin@mysite' => 'My admin name'] style email configuration
+     */
+    private function extractEmailAddress(string|array $email): string
+    {
+        if (is_array($email)) {
+            if (count($email) === 0) {
+                throw new InvalidArgumentException('Empty array passed for $email');
+            }
+            return array_keys($email)[0];
```
ChloeSartorelli, line 113: This will essentially just throw away the "name" part of the email address, which is not the expected behaviour.

GuySartorelli (changes requested):

src/Services/EmailService.php
emteknetnz: Having a look through https://github.com/silverstripe/silverstripe-framework/blob/5/src/Control/Email/Email.php - looks like we may have accidentally lost support for the "Name" portion of array config during the upgrade to symfony/mailer e.g. https://github.com/silverstripe/silverstripe-framework/blob/5/src/Control/Email/Email.php#L206 will also remove the "Name" portion

Given that we should probably add support back in to Email rather than fixing this here

PR #397, opened 2023-02-15
FIX Broken builds

sabina-talipova:
### Description
Draft version of PR. Looks like `InjectionCreator::create` doesn't work as expected. But, probably it's an issue in my local environment.

### Parent issue
- https://github.com/silverstripeltd/product-issues/issues/676

### Broken build
- https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/4121094495/jobs/7116461834

### DIFF
--- a/src/Extensions/ScheduledExecutionExtension.php
+++ b/src/Extensions/ScheduledExecutionExtension.php
@@ -128,8 +128,8 @@ class ScheduledExecutionExtension extends DataExtension
                     $time = DBDatetime::create()->setValue($this->owner->FirstExecution)->Rfc2822();
                 }
 
-                $this->owner->ScheduledJobID = QueuedJobService::singleton()
-                    ->queueJob($job, $time);
+                $queuedJobService = new QueuedJobService();
+                $this->owner->ScheduledJobID = $queuedJobService->queueJob($job, $time);
             }
         }
     }
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -304,6 +304,10 @@ class QueuedJobService
      */
     public function startJob($jobDescriptor, $startAfter = null)
     {
+        if (!$this->queueHandler) {
+            return;
+        }
+
         if ($startAfter && strtotime($startAfter ?? '') > DBDatetime::now()->getTimestamp()) {
             $this->queueHandler->scheduleJob($jobDescriptor, $startAfter);
         } else {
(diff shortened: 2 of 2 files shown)


src/Services/QueuedJobService.php
```diff
@@ -304,6 +304,10 @@ public function queueJob(QueuedJob $job, $startAfter = null, $userId = null, $qu
      */
     public function startJob($jobDescriptor, $startAfter = null)
     {
+        if (!$this->queueHandler) {
```
emteknetnz, line 307: We shouldn't be doing this, a hard exception here is better if `$this->queueHandler` is falsey

`$this->queueHandler` is set via Injector as specified in [yml config](https://github.com/symbiote/silverstripe-queuedjobs/blob/4/_config/queuedjobs.yml#L12) - , by default it is [configured](https://github.com/symbiote/silverstripe-queuedjobs/blob/4/_config/queuedjobs.yml#L6) to use DefaultQueueHandler

src/Extensions/ScheduledExecutionExtension.php
```diff
@@ -128,8 +128,8 @@ public function onBeforeWrite()
                     $time = DBDatetime::create()->setValue($this->owner->FirstExecution)->Rfc2822();
                 }
 
-                $this->owner->ScheduledJobID = QueuedJobService::singleton()
-                    ->queueJob($job, $time);
+                $queuedJobService = new QueuedJobService();
```
emteknetnz, line 131: We shouldn't be changing `::create()` to `new`

src/Services/QueuedJobService.php
```diff
@@ -603,10 +607,12 @@ public function checkDefaultJobs($queue = null)
...
                         }
-                        QueuedJobService::singleton()->queueJob(
-                            Injector::inst()->createWithArgs($jobConfig['type'], $jobConfig['construct']),
-                            date($jobConfig['startDateFormat'] ?? '', strtotime($jobConfig['startTimeString'] ?? ''))
+                        $queuedJob = Injector::inst()->createWithArgs($jobConfig['type'], $jobConfig['construct']);
+                        $queuDate = date(
+                            $jobConfig['startDateFormat'] ?? '',
+                            strtotime($jobConfig['startTimeString'] ?? '')
                         );
+                        $this->queueJob($queuedJob, $queuDate);
```
emteknetnz, line 615: We shouldn't be doing this - we should be using the singleton which presumbably is being used as the central class to keep a record of all the data

src/Services/QueuedJobService.php
```diff
@@ -1516,7 +1522,7 @@ private function addJobHandlersToLogger(LoggerInterface $logger, QueuedJob $job,
             if (!$exists) {
                 // Add the handler
                 /** @var QueuedJobHandler $queuedJobHandler */
-                $queuedJobHandler = QueuedJobHandler::create($job, $jobDescriptor);
+                $queuedJobHandler = new QueuedJobHandler($job, $jobDescriptor);
```
emteknetnz, line 1525: We shouldn't be changing `::create()` to `new`

emteknetnz (changes requested):
Just a note - this PR isn't meant to be merged in its current state

sabina-talipova:
Close as irrelevant to fixing issue.

PR #398, opened 2023-02-15
MNT Ensure parent::__construct() is not passed the wrong param types

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/676

The failures in recipe-kitchen-sink only showed up because
a) we added staticpublishqueue to the sink, and it has this yml - https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/6.0/_config/staticpublishqueue.yml#L24 which leads to this being called - https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/6.0/src/Dev/StaticPublisherState.php#L21
b) the latest version of monolog now has strong typing, previously it was loosely typed - https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/AbstractHandler.php#L36

### DIFF
--- a/tests/QueuedJobsTest/QueuedJobsTest_Handler.php
+++ b/tests/QueuedJobsTest/QueuedJobsTest_Handler.php
@@ -11,6 +11,17 @@ use SilverStripe\Dev\TestOnly;
  */
 class QueuedJobsTest_Handler extends AbstractProcessingHandler implements TestOnly
 {
+    public function __construct()
+    {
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):

GuySartorelli:
Short version of the comment: We pass arguments in when instantiating a new one of these via Injector. The new constructor we've added here effectively says "Ignore the arguments, we don't want them, they're inconsistent with the parent constructor."

PR #404, opened 2023-04-27
FIX Broken TaskLink & QueueLink

sabina-talipova:
## Description
String concatenation for getting link was replaced with `Controller::join_links(...)` .  

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10757

### DIFF
--- a/src/Controllers/QueuedTaskRunner.php
+++ b/src/Controllers/QueuedTaskRunner.php
@@ -3,6 +3,7 @@
 namespace Symbiote\QueuedJobs\Controllers;
 
 use SilverStripe\Admin\AdminRootController;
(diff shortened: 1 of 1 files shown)


GuySartorelli (changes requested):
There are other inappropriate string concatenations of URLs in this file (e.g. lines 191, 192). Please clean those up as well, and have a quick glance over the rest of the module just in case it happens elsewhere in the module as well.

sabina-talipova:
> have a quick glance over the rest of the module just in case it happens elsewhere in the module as well

DONE.
Didn't identified another places in the code where we concatenate string for getting link.

GuySartorelli (approved):
Nice, works well.

PR #421, opened 2024-04-24
FIX Replace assignment with intended comparison

oddnoc:
## Expected behaviour

If `E_DEPRECATED` is defined and `$errno` is either `E_DEPRECATED` or `E_USER_DEPRECATED`, return from `handleError()`.

## Actual behaviour

If `E_DEPRECATED` is defined and `$errno` is any value, return from `handleError()`. This means that exceptions are never thrown by the method whenever `E_DEPRECATED` is defined.

### DIFF
--- a/src/Services/JobErrorHandler.php
+++ b/src/Services/JobErrorHandler.php
@@ -27,7 +27,7 @@ class JobErrorHandler
         if (error_reporting()) {
             // Don't throw E_DEPRECATED in PHP 5.3+
             if (defined('E_DEPRECATED')) {
-                if ($errno == E_DEPRECATED || $errno = E_USER_DEPRECATED) {
+                if (E_DEPRECATED == $errno || E_USER_DEPRECATED == $errno) {
                     return;
                 }
             }


src/Services/JobErrorHandler.php
```diff
@@ -27,7 +27,7 @@ public function handleError($errno, $errstr, $errfile, $errline)
         if (error_reporting()) {
             // Don't throw E_DEPRECATED in PHP 5.3+
             if (defined('E_DEPRECATED')) {
-                if ($errno == E_DEPRECATED || $errno = E_USER_DEPRECATED) {
+                if (E_DEPRECATED == $errno || E_USER_DEPRECATED == $errno) {
```
ChloeSartorelli, line 30: Please don't use YODA syntax.

GuySartorelli (changes requested):

GuySartorelli:
Please also retarget this PR to the `5.1` branch so it can be released as a patch.

oddnoc:
Will open a new PR

PR #422, opened 2024-04-26
Replace assignment with intended comparison

oddnoc:
Correct inadvertent assignment typo that results in unreachable code.


GuySartorelli (approved):
LGTM, thanks for submitting this.

PR #436, opened 2024-06-24
NEW Send an email notification when a job is broken

emteknetnz:
Issue https://github.com/symbiote/silverstripe-queuedjobs/issues/435

There is an email sent for 'stalled' jobs, however there is not an email sent for 'broken' jobs. There is existing logic to ensure that there is only a single notification sent, however no email is actually sent, instead the only notification goes to the logger. An [inline comment](https://github.com/symbiote/silverstripe-queuedjobs/blob/4/src/Services/QueuedJobService.php#L500) says "email" so looks like the intention all along was to send an email.

Testing steps

Create a vanilla install of silverstripe/installer + this PR

**_config/myemail.yml**
```yml
---
Name: myemail
After:
  - '#project-emailconfig'
---
SilverStripe\Control\Email\Email:
  admin_email:
    admin-email@example.com: 'Admin-email'
# + something to setup mailhog / mailcatcher / etc
```

**MyBrokenJob.php**
```php
<?php

use Symbiote\QueuedJobs\Services\AbstractQueuedJob;
use Symbiote\QueuedJobs\Services\QueuedJob;

class MyBrokenJob extends AbstractQueuedJob
{
    public function getTitle()
    {
        return 'My broken job';
    }

    public function getJobType() {
        return QueuedJob::QUEUED;
    }

    public function process()
    {
        throw new Exception('MY BROKEN JOB EXCEPTION');
    }
}
```

In `admin/queuedjobs` create a new 'My broken job'

In console:
`vendor/bin/sake dev/tasks/ProcessJobQueueTask`
`vendor/bin/sake dev/tasks/CheckJobHealthTask`

There should be an 'Broken job' email that was sent

### DIFF
--- a/src/Services/EmailService.php
+++ b/src/Services/EmailService.php
@@ -6,7 +6,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\Email\Email;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injectable;
-use SilverStripe\Subsites\Model\Subsite;
+use SilverStripe\ORM\DataList;
 
 /**
  * Class EmailService
@@ -35,11 +35,6 @@ class EmailService
         );
     }
 
-    /**
-     * @param array $jobConfig
-     * @param string $title
-     * @return Email|null
-     */
     public function createMissingDefaultJobReport(array $jobConfig, string $title): ?Email
     {
         $subject = sprintf('Default Job "%s" missing', $title);
@@ -59,12 +54,6 @@ class EmailService
             ->setHTMLTemplate('QueuedJobsDefaultJob');
     }
 
-    /**
-     * @param string $subject
-     * @param string $message
-     * @param int $jobID
-     * @return Email|null
-     */
     public function createStalledJobReport(string $subject, string $message, int $jobID): ?Email
     {
         $email = $this->createReport($subject);
@@ -81,12 +70,24 @@ class EmailService
             ->setHTMLTemplate('QueuedJobsStalledJob');
     }
 
+    public function createBrokenJobsReport(string $subject, string $message, DataList $jobs): ?Email
+    {
+        $email = $this->createReport($subject);
+        if ($email === null) {
+            return null;
+        }
+        return $email
+            ->setData([
+                'Message' => $message,
+                'Jobs' => $jobs,
+                'Site' => Director::absoluteBaseURL(),
+            ])
+            ->setHTMLTemplate('QueuedJobsBrokenJobs');
+    }
+
     /**
      * Create a generic email report
      * useful for reporting queue service issues
-     *
-     * @param string $subject
-     * @return Email|null
      */
     public function createReport(string $subject): ?Email
     {
--- a/src/Services/QueuedJobService.php
+++ b/src/Services/QueuedJobService.php
@@ -519,6 +519,15 @@ class QueuedJobService
             ]
         );
 
+        // Send broken job email notification
+        $subject = _t(__CLASS__ . '.BROKEN_JOBS', 'Broken job(s)');
+        $message = _t(__CLASS__ . '.BROKEN_JOBS_MSG', 'The following job(s) appear to be broken.');
+        $email = EmailService::singleton()->createBrokenJobsReport($subject, $message, $brokenJobs);
+        if ($email) {
+            $email->send();
+        }
+
+        // Mark broken jobs as notified
         $placeholders = implode(', ', array_fill(0, count($brokenIDs ?? []), '?'));
         $query = SQLUpdate::create(
             '"QueuedJobDescriptor"',
--- a/templates/QueuedJobsBrokenJobs.ss
+++ b/templates/QueuedJobsBrokenJobs.ss
@@ -1,0 +1,7 @@
+$Message
+
+<% loop $Jobs %>
+- $JobTitle (ID: $ID)
+<% end_loop %>
+
+Log in to $Site to see further details and take any necessary actions.
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -19,6 +19,8 @@ en:
     PLURALNAME: 'Queued Job Rules'
     SINGULARNAME: 'Queued Job Rule'
   QueuedJobs:
+    BROKEN_JOBS: 'Broken job(s)'
+    BROKEN_JOBS_MSG: 'The following job(s) appear to be broken.'
     CREATE_JOB_TYPE: 'Create job of type'
     CREATE_NEW_JOB: 'Create new job'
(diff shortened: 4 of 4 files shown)


src/Services/QueuedJobService.php
```diff
@@ -519,6 +519,25 @@ public function checkJobHealth($queue = null)
             ]
         );
 
+        // Send broken job email notification
+        $subject = _t(__CLASS__ . '.BROKEN_JOB', 'Broken job');
+        $message = sprintf(_t(
```
emteknetnz: Using a combination of sprintf() with _t() is pretty standard elsewhere in the codebase.

You can used named variables with _t(), though that's much less common

src/Services/EmailService.php
```diff
@@ -35,11 +34,6 @@ public function __construct()
         );
     }
 
-    /**
-     * @param array $jobConfig
```
emteknetnz, line 39: Unrelated refactoring, methods are strongly typed do not need docblock types

src/Services/EmailService.php
```diff
@@ -6,7 +6,6 @@
 use SilverStripe\Control\Email\Email;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injectable;
-use SilverStripe\Subsites\Model\Subsite;
```
emteknetnz, line 9: Unrelated refactoring, unused import

src/Services/QueuedJobService.php
```diff
@@ -519,6 +519,24 @@ public function checkJobHealth($queue = null)
...
+                'The following job(s) appear to be broken.',
+                'Log in to %s to see further details and take any necessary actions.',
+            ])
+        ), Director::absoluteBaseURL());
+        foreach ($brokenJobs as $brokenJob) {
+            $message .= sprintf("\n- %s (ID: %s)", $brokenJob->JobTitle, $brokenJob->ID);
+        }
+        $email = EmailService::singleton()->createBrokenJobReport($subject, $message);
+        if ($email) {
+            $email->sendPlain();
```
ChloeSartorelli: Why specifically `sendPlain()`? The other emails in this file use `send()`.

src/Services/QueuedJobService.php
```diff
@@ -519,6 +519,24 @@ public function checkJobHealth($queue = null)
...
+        $message = sprintf(_t(
+            __CLASS__ . '.BROKEN_JOBS_MSG',
+            implode(' ', [
+                'The following job(s) appear to be broken.',
+                'Log in to %s to see further details and take any necessary actions.',
+            ])
+        ), Director::absoluteBaseURL());
+        foreach ($brokenJobs as $brokenJob) {
+            $message .= sprintf("\n- %s (ID: %s)", $brokenJob->JobTitle, $brokenJob->ID);
+        }
```
ChloeSartorelli: Probably better to use an email template, same as the other emails sent from this module.

GuySartorelli (changes requested):
Works well locally, just a couple of small things to fix.

src/Services/QueuedJobService.php
emteknetnz: Updated

src/Services/QueuedJobService.php
emteknetnz: Updated

templates/QueuedJobsBrokenJobs.ss
```diff
@@ -0,0 +1 @@
+$Message.RAW
```
ChloeSartorelli: Should be more like the `QueuedJobsStalledJob.ss` file.

```suggestion
Hi,

$Message

<% loop $Jobs $>
- $JobTitle (ID: $ID)
<% end_loop %>

Log in to $Site to see further details and take any necessary actions.
```

src/Services/QueuedJobService.php
```diff
@@ -519,6 +519,26 @@ public function checkJobHealth($queue = null)
...
+            implode(' ', [
+                'The following job(s) appear to be broken.',
+                'Log in to %s to see further details and take any necessary actions.',
+            ])
+        ), Director::absoluteBaseURL()));
+        foreach ($brokenJobs as $brokenJob) {
+            // Using htmlentities() because of the use of $Message.RAW in QueuedJobsBrokenJobs.ss
+            // in order for the <br> tags to render as HTML
+            $message .= sprintf("<br>- %s (ID: %s)", htmlentities($brokenJob->JobTitle), $brokenJob->ID);
+        }
```
ChloeSartorelli:
```suggestion
        $message = 'The following job(s) appear to be broken.';
```
You can use `_t` if you want - the point is it should just be the message part, not the full template.

src/Services/EmailService.php
```diff
@@ -81,12 +69,22 @@ public function createStalledJobReport(string $subject, string $message, int $jo
...
 
+    public function createBrokenJobsReport(string $subject, string $message): ?Email
+    {
+        $email = $this->createReport($subject);
+        if ($email === null) {
+            return null;
+        }
+        return $email
+            ->setData([
+                'Message' => $message,
```
ChloeSartorelli, line 81: Pass in the jobs and site as well, as per suggested change to template.

GuySartorelli (changes requested):

src/Services/QueuedJobService.php
emteknetnz: Updated - though I've retained most of the message here so that's all translatable

src/Services/EmailService.php
emteknetnz: Updated

src/Services/QueuedJobService.php
```diff
@@ -519,6 +519,21 @@ public function checkJobHealth($queue = null)
             ]
         );
 
+        // Send broken job email notification
+        $subject = _t(__CLASS__ . '.BROKEN_JOBS', 'Broken job(s)');
+        $message = sprintf(_t(
+            __CLASS__ . '.BROKEN_JOBS_MSG',
+            implode(' ', [
+                'The following job(s) appear to be broken.',
+                'Log in to %s to see further details and take any necessary actions.',
```
ChloeSartorelli: The other templates have the "log in to etc etc" part in the template. We should aim to have this template be similar to the existing templates.

If you really want to use i18n on it, you should use the template `<%t %>` and update the other templates to also use it. See https://docs.silverstripe.org/en/5/developer_guides/i18n/#usage-in-template-files for docs on how to use i18n in templates.

GuySartorelli (changes requested):

src/Services/QueuedJobService.php
emteknetnz: I don't feel strongly about it, I've updated it to make it similar to the other emails

GuySartorelli (approved):
LGTM

PR #447, opened 2024-09-10
API Deprecate API that will be removed

GuySartorelli:
See https://github.com/symbiote/silverstripe-queuedjobs/pull/445

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11341

### DIFF
--- a/src/Services/AbstractQueuedJob.php
+++ b/src/Services/AbstractQueuedJob.php
@@ -271,4 +271,31 @@ abstract class AbstractQueuedJob implements QueuedJob, UserContextInterface
     {
         return isset($this->jobData->$name) ? $this->jobData->$name : null;
     }
+
+    /**
+     * Resolves a queue name to one of the queue constants.
+     * If $queue is already the value of one of the constants, it will be returned.
+     * If the queue is unknown, `null` will be returned.
+     */
+    public static function getQueue(string|int $queue): ?string
+    {
+        switch (strtolower($queue)) {
+            case 'immediate':
+                $queue = QueuedJob::IMMEDIATE;
+                break;
+            case 'queued':
+                $queue = QueuedJob::QUEUED;
+                break;
+            case 'large':
+                $queue = QueuedJob::LARGE;
+                break;
+            default:
+                $queue = (string) $queue;
+                $queues = [QueuedJob::IMMEDIATE, QueuedJob::QUEUED, QueuedJob::LARGE];
+                if (!ctype_digit($queue) || !in_array($queue, $queues)) {
+                    return null;
+                }
+        }
+        return $queue;
+    }
 }
--- a/src/Tasks/ProcessJobQueueTask.php
+++ b/src/Tasks/ProcessJobQueueTask.php
@@ -8,6 +8,8 @@ use Monolog\Logger;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\Environment;
 use SilverStripe\Dev\BuildTask;
+use SilverStripe\Dev\Deprecation;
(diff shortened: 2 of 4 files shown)


src/Services/AbstractQueuedJob.php
```diff
@@ -271,4 +271,31 @@ public function __get($name)
     {
         return isset($this->jobData->$name) ? $this->jobData->$name : null;
     }
+
+    /**
+     * Resolves a queue name to one of the queue constants.
+     * If $queue is already the value of one of the constants, it will be returned.
+     * If the queue is unknown, `null` will be returned.
+     */
+    public static function getQueue(string|int $queue): ?string
```
ChloeSartorelli, line 280: We need this in two places so I've abstracted it out from the task into here.

src/Services/AbstractQueuedJob.php
```diff
@@ -271,4 +271,31 @@ public function __get($name)
...
+                $queue = QueuedJob::LARGE;
+                break;
+            default:
+                $queue = (string) $queue;
+                $queues = [QueuedJob::IMMEDIATE, QueuedJob::QUEUED, QueuedJob::LARGE];
+                if (!ctype_digit($queue) || !in_array($queue, $queues)) {
+                    return null;
+                }
+        }
+        return $queue;
```
emteknetnz, line 299:
```suggestion
        if (is_int($queue)) {
            return $queue;
        }
        return match(strtolower($queue)) {
            'immediate' => QueuedJob::IMMEDIATE,
            'queued' => QueuedJob::QUEUED,
            'large' => QueuedJob::LARGE,
            default => null,
        }
```

Not sure of the value of having string|int as a param is, string only would be better.

Method name might benefit from being called something like getQueueFromType

src/Tasks/ProcessJobQueueTask.php
```diff
@@ -83,7 +85,7 @@ public function run($request)
         }
 
         // Run the queue
-        $queue = $this->getQueue($request);
+        $queue = AbstractQueuedJob::getQueue($request->getVar('queue') ?? 'Queued');
```
emteknetnz, line 88:
```suggestion
        $queue = AbstractQueuedJob::getQueue($request->getVar('queue') ?? 'queued');
```

emteknetnz (changes requested):

src/Services/AbstractQueuedJob.php
ChloeSartorelli: `int` is there because the consts are string values of ints (`"0"`, `"1"`, etc. So it makes sense to allow someone to pass in `1` and get the corresponding `"1"` for `===` matching against the const.

ChloeSartorelli: I'm not sure how `getQueueFromType` is better.... maybe `parseValueToQueueConst` or something like that, if you really want it renamed?

Ultimately though I can't see anyone actually using this, I only moved it because _we_ now use it in more than one place.

src/Tasks/ProcessJobQueueTask.php
ChloeSartorelli: `Queued` is taken from the original method I'm moving this from. I'm hesitant to change it as it's not clear if it was like that for a good reason. Though I guess it goes through `strtolower` so it doesn't matter whether it's caps or not. I'd prefer to just leave it.

src/Services/AbstractQueuedJob.php
emteknetnz: All good, I'm not too worried about the rename given it's the same name as the old method

I've approved all the other PRs at this point and it's unit tested so don't worry about recommended code refactor

emteknetnz (approved):

PR #457, opened 2025-01-06
FIX Stop using deprecated API

GuySartorelli:
There are two commits here - feel free to squash if you want but can be left separate.

## Issue
- https://github.com/silverstripe/.github/issues/311

### DIFF
--- a/src/Jobs/DoormanQueuedJobTask.php
+++ b/src/Jobs/DoormanQueuedJobTask.php
@@ -2,7 +2,6 @@
 
 namespace Symbiote\QueuedJobs\Jobs;
 
-use SilverStripe\Dev\Deprecation;
 use AsyncPHP\Doorman\Cancellable;
 use AsyncPHP\Doorman\Expires;
 use AsyncPHP\Doorman\Process;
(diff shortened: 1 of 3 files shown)


src/Jobs/DoormanQueuedJobTask.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Symbiote\QueuedJobs\Jobs;
 
-use SilverStripe\Dev\Deprecation;
```
ChloeSartorelli, line 5: Not used

src/Services/JobErrorHandler.php
```diff
@@ -25,14 +25,9 @@ public function clear()
...
-            // Don't throw E_DEPRECATED in PHP 5.3+
-            if (defined('E_DEPRECATED')) {
-                if ($errno == E_DEPRECATED || $errno == E_USER_DEPRECATED) {
-                    return;
-                }
-            }
-
             switch ($errno) {
+                case E_DEPRECATED:
+                case E_USER_DEPRECATED:
```
ChloeSartorelli, line 30: Not really related but noticed it and seemed sensible to update this while I'm here, since `E_DEPRECATED` will always be defined in supported PHP versions now.

emteknetnz (approved):

PR #474, opened 2025-06-04
ENH: CMS edit link for queued job descriptor.

mfendeksilverstripe:
# CMS edit link for queued job descriptor

Added the ability for Queued job descriptor to generate its own CMS edit link to support scenarios where this link needs to be generate in the CMS UI after some CMS action created a job.

## Related issues

* https://github.com/silverstripe/silverstripe-queuedjobs/issues/475

### DIFF
--- a/_config/queuedjobs.yml
+++ b/_config/queuedjobs.yml
@@ -30,6 +30,11 @@ SilverStripe\SiteConfig\SiteConfig:
   extensions:
     - Symbiote\QueuedJobs\Extensions\MaintenanceLockExtension
 
+Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor:
+  cms_edit_owner: Symbiote\QueuedJobs\Controllers\QueuedJobsAdmin
+  extensions:
+    - SilverStripe\Admin\CMSEditLinkExtension
+
 ---
 Name: gearman_queue_settings
 Only:
--- a/src/Controllers/QueuedJobsAdmin.php
+++ b/src/Controllers/QueuedJobsAdmin.php
@@ -47,11 +47,11 @@ class QueuedJobsAdmin extends ModelAdmin
      */
     private static $menu_icon_class = 'font-icon-checklist';
 
-    /**
-     * @var array
-     */
-    private static $managed_models = [
-        QueuedJobDescriptor::class
+    private static array $managed_models = [
+        'jobs' => [
+            'title' => 'Jobs',
+            'dataClass' => QueuedJobDescriptor::class,
+        ]
     ];
 
     /**


src/Controllers/QueuedJobsAdmin.php
```diff
@@ -47,11 +47,11 @@ class QueuedJobsAdmin extends ModelAdmin
      */
     private static $menu_icon_class = 'font-icon-checklist';
 
-    /**
-     * @var array
-     */
-    private static $managed_models = [
-        QueuedJobDescriptor::class
+    private static array $managed_models = [
+        'jobs' => [
```
mfendeksilverstripe, line 51: Noticed that the CMS edit link which is generated without this change doesn't work. Looks like this segment causes an issue:

`Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor`

it should be `QueuedJobDescriptor`

Leaving this in as a good workaround as this is probably a good change to make on its own.

_config/queuedjobs.yml
```diff
@@ -30,6 +30,11 @@ SilverStripe\SiteConfig\SiteConfig:
   extensions:
     - Symbiote\QueuedJobs\Extensions\MaintenanceLockExtension
 
+Symbiote\QueuedJobs\DataObjects\QueuedJobDescriptor:
+  cms_edit_owner: Symbiote\QueuedJobs\Controllers\QueuedJobsAdmin
+  extensions:
+    - SilverStripe\Admin\CMSEditLinkExtension
```
ChloeSartorelli, line 36: Please add these as `private static` properties directly in the class itself.

src/Controllers/QueuedJobsAdmin.php
ChloeSartorelli: We should ideally not be changing the URL segment. It sounds like you've found a bug with the way `CMSEditLinkExtension` works - that bug should be fixed instead of working around it here.

GuySartorelli (changes requested):

GuySartorelli:
Please also target the `6` branch. There will be no new minor releases under the `5` release line.

GuySartorelli:
Also there should be no `:` after `ENH` in the commit message

src/Controllers/QueuedJobsAdmin.php
mfendeksilverstripe: Need some guidance on this one @GuySartorelli . As far as I can see there is some sort of naming convention present in the `GridField` of the `ModelAdmin`.

```php
public function getCMSEditLinkForManagedDataObject(DataObject $obj): string
{
    $modelTab = $this->getModelTabForModelClass($obj->ClassName);
    if ($modelTab === null) {
        throw new InvalidArgumentException("$obj->ClassName isn't managed by this ModelAdmin");
    }
    $link = static::join_links(
        $this->getLinkForModelClass($obj->ClassName),
        'EditForm/field/',
        $this->sanitiseClassName($modelTab),// <--- This is the important bit
        'item',
        $obj->ID
    );
    $this->extend('updateEditLinkForDataObject', $link, $obj);
    return $link;
}

protected function getGridField(): GridField
{
    $field = GridField::create(
        $this->sanitiseClassName($this->modelTab), // <--- This is the important bit
        false,
        $this->getList(),
        $this->getGridFieldConfig()
    );

    $this->extend('updateGridField', $field);

    return $field;
}
```
This module doesn't follow the convention, see `QueuedJobsAdmin`:

```php
$grid = GridField::create(
    'QueuedJobDescriptor',// <--- This is the important bit
    '',
    $list,
    $gridFieldConfig
);
$grid->setForm($form);
/** @skipUpgrade */
$form->Fields()->replaceField($this->sanitiseClassName(QueuedJobDescriptor::class), $grid);
```

So my take on this is either that this module needs to follow the naming convention or needs to define a model tab via the configuration change that I proposed. Please let me know what's your preference.

ChloeSartorelli: Sorry it took so long to get back to this. I think every time I read the question I got confused, so I put it on the shelf for when I had more time to try decipher it :sweat_smile: 

I had to go debug this to understand the question - but now I see the name of the `GridField` in `QueuedJobsAdmin` doesn't match the name `ModelAdmin` uses.

For reference, the URL currently being generated looks like this:
`https://mysite.local/admin/queuedjobs/Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor/EditForm/field/Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor/item/2`
The working URL would look like this:
`https://mysite.local/admin/queuedjobs/Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor/EditForm/field/QueuedJobDescriptor/item/2/`

i.e. the name of the GridField is `QueuedJobDescriptor` but the expected name for the URL generation is `Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor`.

The two solutions I can see (without changing `managed_models` configuration) are to either:
1. Change the name of the GridField (one of your proposed solutions) - which could break some peoples' customisations
2. Implement `getCMSEditLinkForManagedDataObject()` in `QueuedJobsAdmin`.

I prefer the latter, as it has little (if any) risk of breaking existing customisations to the queued jobs admin.
An implementation might look like this:

```php
public function getCMSEditLinkForManagedDataObject(DataObject $obj): string
{
    if ($obj instance of QueuedJobDescriptor) {
        // The name of the GridField doesn't match the standard ModelAdmin naming convention
        // so we have to build this link ourselves
        $link = static::join_links(
            $this->getLinkForModelClass($obj->ClassName),
            'EditForm/field/QueuedJobDescriptor/item',
            $obj->ID
        );
        $this->extend('updateEditLinkForDataObject', $link, $obj);
        return $link;
    }
    // Allow for other managed models
    return parent::getCMSEditLinkForManagedDataObject($obj);
}
```

mfendeksilverstripe: Thanks for feedback, I don't think this is really worth patching as this can already be easily fixed via an extension within project code which renames the field.

I'm keen on making the changes that actually address the root cause which is the non-standard GridField name. Please let me know if I can proceed with a breaking change instead. I understand that this would push the fix back further but my aim is not to fix this as soon as possible.

ChloeSartorelli: We don't have a branch set up for breaking changes, and our process is to not create one until we're ready to start working on breaking changes across the board which won't be until next year most likely. Until then we won't accept any PRs which introduce breaking changes.

mfendeksilverstripe: All good, will close this PR then as I don't plan t spend time on patching this.

mfendeksilverstripe:
Closed due to compatibility constraints - the root cause fix is to rename the GridField which is a breaking change. We should open a new PR once we have the option to do so.

mfendeksilverstripe:
Closed due to compatibility constraints - the root cause fix is to rename the GridField which is a breaking change. We should open a new PR once we have the option to do so.

Issue #484, opened 2025-08-24
CMS fields cleanup (Job descriptor)


mfendeksilverstripe:
# CMS fields cleanup (Job descriptor)

* current CMS UI has many fields, we are lacking structure
* we are missing `beforeUpdateCMSFields` hook to allow easier customisation

## PRs

- https://github.com/silverstripe/silverstripe-queuedjobs/pull/485
- https://github.com/silverstripe/developer-docs/pull/820

GuySartorelli:
Please use the appropriate issue template in the future. Starting from a blank issue is for maintainer use only.

GuySartorelli:
PRs merged

PR #485, opened 2025-08-24
ENH CMS fields cleanup.

mfendeksilverstripe:
# ENH CMS fields cleanup

* `getCMSFields()` wrapped up inside `beforeUpdateCMSFields()` for better customisation
* existing fields grouped logically using `CompositeField` for better CMS UI experience

It's highly recommended to turn off WS when reading the diff.

## Issue

-  https://github.com/silverstripe/silverstripe-queuedjobs/issues/484

### DIFF
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -425,15 +425,14 @@ class QueuedJobDescriptor extends DataObject
             ]);
 
             // Display override to remove the left margin to address a systemic issue with CompositeField
-            // TODO remove this patch after https://github.com/silverstripe/silverstripe-framework/issues/11857 is fixed
             $cssOverrideContent = <<<'HTML'
-<style>
-.composite,
-.composite > .form__field-holder {
-  margin-left:0 !important;
-}
-</style>
-HTML;
+            <style>
+            .composite,
+            .composite > .form__field-holder {
+              margin-left:0 !important;
+            }
+            </style>
+            HTML;
 
             $progressBarContent = sprintf(
                 '<p>%3$0.2f%% completed</p><p><progress value="%1$d" max="%2$d">%3$0.2f%%</progress></p>',
--- a/src/DataObjects/QueuedJobDescriptor.php
+++ b/src/DataObjects/QueuedJobDescriptor.php
@@ -424,14 +424,13 @@ class QueuedJobDescriptor extends DataObject
                 'WorkerCount',
             ]);
 
-//            public const string TYPE_GOOD = 'good'; // green
-//            public const string TYPE_NOTICE = 'notice'; // blue
-//            public const string TYPE_WARNING = 'warning'; //orange
-//            public const string TYPE_ERROR = 'error'; //red
-//            public const string TYPE_PLAIN = ''; // transparent
-//
-//            public const string CATEGORY_MESSAGE = 'message';
-//            public const string CATEGORY_ALERT = 'alert';
+            $progressBarContent = sprintf(
+                '<p>%3$0.2f%% completed</p><p><progress value="%1$d" max="%2$d">%3$0.2f%%</progress></p>',
+                $this->StepsProcessed,
+                $this->TotalSteps,
+                $this->TotalSteps > 0 ? ($this->StepsProcessed / $this->TotalSteps) * 100 : 0
+            );
+
             $timelineHeadingContent = <<<'HTML'
 <p class="alert warning">
 It is recommended to avoid editing these fields as they are managed by the Queue Runner / Service.
@@ -441,15 +440,7 @@ HTML;
             $fields->addFieldsToTab('Root.Main', [
                 CompositeField::create([
                     HeaderField::create('ProgressReportHeading', 'Progress'),
-                    LiteralField::create(
-                        'JobProgressReportIntro',
-                        sprintf(
-                            '<p>%3$0.2f%% completed</p><p><progress value="%1$d" max="%2$d">%3$0.2f%%</progress></p>',
-                            $this->StepsProcessed,
-                            $this->TotalSteps,
-                            $this->TotalSteps > 0 ? ($this->StepsProcessed / $this->TotalSteps) * 100 : 0
-                        )
-                    ),
+                    LiteralField::create('JobProgressReportIntro', $progressBarContent),
                 ]),
                 CompositeField::create([
                     HeaderField::create('OverviewHeading', 'Overview'),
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Please use the PR template in the future.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -389,188 +389,214 @@ public function getJobStatusValues(): array
         ];
     }
 
-    /**
-     * @return FieldList
-     */
-    public function getCMSFields()
+    public function getCMSFields(): FieldList
```
ChloeSartorelli: We can't update the method signature outside of a major release.

```suggestion
    /**
     * @return FieldList
     */
    public function getCMSFields()
```

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -389,188 +389,214 @@ public function getJobStatusValues(): array
...
+                $fields->remove($field);
+            }
 
-        if (Permission::check('ADMIN')) {
-            return $fields;
-        }
+            // Add read only fields
+            foreach ($readonlyList as $field) {
+                $fields->add($field);
+            }
```
ChloeSartorelli: Why are you doing this?

Seems like instead we should leave this until the end of `getCMSFields()` since we presumably want _all_ fields to be readonly, not only the ones from before the extension hooks.

So the method should end with `return parent::getCMSFields()->makeReadonly();`

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -389,188 +389,214 @@ public function getJobStatusValues(): array
...
+            ]);
 
-        $startAfter->setDescription(
-            'Used to prevent the job from starting earlier than the specified time.'
-            . ' Note that this does not guarantee that the job will start'
-            . ' exactly at the specified time (it will start the next time the cron job runs).'
-        );
+            // Main
+            $fields->addFieldsToTab('Root.Main', [
+                CompositeField::create([
```
ChloeSartorelli, line 451: Missing `use` statement for `CompositeField`

GuySartorelli (changes requested):
## Composite fields

I think this does overall look cleaner with the composite fields, but it also adds a bunch of blank space to the left of the form.

This is especially bad on certain widths where the space is _only_ on the left, not on the right.

If you want to keep these, the queued jobs form probably needs some custom CSS (or add an appropriate css class like "ms-0" via `addExtraClass()` to the composite fields) to remove the left margin.

**Before this PR**
Screenshot from 2025-08-25 15-36-44

**With this PR**

Screenshot from 2025-08-25 15-36-55

## Changelog

It's possible this will cause problems with existing `updateCMSFields()` extension hook implementations for this model or subclasses of it.

Please add a one-liner bullet point to the "other new features and enhancements" section of the CMS 6.2 changelog (after https://github.com/silverstripe/developer-docs/pull/804 has been merged).

mfendeksilverstripe:
@GuySartorelli I've pushed up new changes as per your feedback.

I'm reluctant to customise the margins as the layout on the main tab is currently consistent with the advanced tab. As far as I can see this is the expected display of the form fields using `CompositeField` and `FieldGroup`.

Please review and let me know if further changes are needed.

GuySartorelli:
> I'm reluctant to customise the margins as the layout on the main tab is currently consistent with the advanced tab.

Sorry, I'm not sure what you mean by this? Without this PR, both the "Main" and "Advanced" tabs have labels on the left, fields in the middle. This PR adds empty space to the left in both tabs. I'm asking that you please remove that space from the left in both tabs - which will retain consistency.

The space on the left is fine at large screen widths since it's balanced on the right, but at lower screen widths you have _only_ whitespace on the left and not on the right, which looks quite bad. I'd be okay if you kept the space at large screen widths and only removed it for smaller screen widths. The breakpoint is at 1200px

GuySartorelli:
Please also remember to create a changelog PR as requested in https://github.com/silverstripe/silverstripe-queuedjobs/pull/485#pullrequestreview-3149735236

mfendeksilverstripe:
> Sorry, I'm not sure what you mean by this?

What I'm trying to point out is that the current layout CMS fields for JobDescriptor is using standard set of form fields without any CSS customisation and I would prefer to keep it that way to keep the baseline solution simple and extensible.

If there is an issue with left margin of `CompositeField` and `FieldGroup` it should be addressed as a separate issue so the fix can be applied to all instances of these fields not just this particular scenario.

I will raise a changelog PR once this change-set is solidified.

GuySartorelli:
Usually these composite fields or field groups are used sparingly in a CMS field list, not wrapping all of the fields, so the left margin is usually appropriate.

In this context the left margin is not appropriate. It's specific to the way you're using these classes to wrap _all_ of the fields in these tabs rather than an inherent problem with the styling of these elements in general use.

For this specific context, please remove the left margin at least for widths less than 1200px.

mfendeksilverstripe:
@GuySartorelli We actually use `CompositeField` and `FieldGroup` a lot in my projects. Are there any docs related to the intended use of form fields?

I still think that there is value in having this fix be global is a sense that it might be a display option for these fields as opposed to limiting this to only this feature.

GuySartorelli:
If you're willing for this improvement to be on hold until we prioritise a global UX review on `CompositeField` and `FieldGroup` to validate whether a global change is appropriate then I guess we can do that. I can't promise when (or even if) that would be prioritised though so I wouldn't recommend that approach.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -424,6 +424,17 @@ public function getCMSFields()
                 'WorkerCount',
             ]);
 
+            // Display override to remove the left margin to address a systemic issue with CompositeField
+            // TODO remove this patch after https://github.com/silverstripe/silverstripe-framework/issues/11857 is fixed
```
ChloeSartorelli:
```suggestion
```
Please don't include TODO comments in the code. Instead, the issue you raised can refer to this code. That way when that issue is processed, if we do decide there's a systemic problem, we can remove this code at the same time as resolving the systemic problem.

src/DataObjects/QueuedJobDescriptor.php
```diff
@@ -424,6 +424,17 @@ public function getCMSFields()
...
+            $cssOverrideContent = <<<'HTML'
+<style>
+.composite,
+.composite > .form__field-holder {
+  margin-left:0 !important;
+}
+</style>
+HTML;
```
ChloeSartorelli:
```suggestion
            <style>
            .composite,
            .composite > .form__field-holder {
              margin-left:0 !important;
            }
            </style>
            HTML;
```
You can keep this indented in modern PHP.

GuySartorelli (changes requested):

src/DataObjects/QueuedJobDescriptor.php
mfendeksilverstripe: Issue raised https://github.com/silverstripe/silverstripe-framework/issues/11857

mfendeksilverstripe:
Have pushed up the requested changes @GuySartorelli , please review.

mfendeksilverstripe:
@GuySartorelli Changelog entry added in https://github.com/silverstripe/developer-docs/pull/820

GuySartorelli (approved):
Looks good! Thanks for submitting this and being patient through the review process.

PR #487, opened 2025-11-26
ENH PHP 8.5 Support

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/455

The following changes were made based on the [PHP 8.5 changelog](https://www.php.net/ChangeLog-8.php):

- The setAccessible() methods of various Reflection objects have been deprecated, as those no longer have an effect.
- Non-canonical cast names (boolean), (integer), (double), and (binary) have been deprecated.

### DIFF
--- a/src/QJUtils.php
+++ b/src/QJUtils.php
@@ -74,8 +74,6 @@ class QJUtils
             $val = 'NULL';
         } elseif (is_int($val)) {
             $val = (int) $val;
-        } elseif (is_double($val)) {
-            $val = (double) $val;
         } elseif (is_float($val)) {
             $val = (float) $val;
         } else {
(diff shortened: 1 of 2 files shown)


src/QJUtils.php
```diff
@@ -74,9 +74,7 @@ protected function recursiveQuote($val)
             $val = 'NULL';
         } elseif (is_int($val)) {
             $val = (int) $val;
-        } elseif (is_double($val)) {
-            $val = (double) $val;
-        } elseif (is_float($val)) {
+        } elseif (is_float($val) || is_double($val)) {
```
ChloeSartorelli: https://www.php.net/manual/en/function.is-double.php
> is_double — Alias of [is_float()](https://www.php.net/manual/en/function.is-float.php)
```suggestion
        } elseif (is_float($val)) {
```

GuySartorelli (changes requested):

src/QJUtils.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #495, opened 2026-07-22
job option no longer works as expected via CLI


torleif:
### Module version(s) affected

^6.0

### Description

in the latest version, if you want to run a specific job you used to be able to add an option to specify a job descriptor ID. however there's a bug in the code that disables the numeric id from working as expected:

**ProcessJobQueueTask.php**
```

        // Check if there is a job to run
        $job = $input->getOption('job');
        if ($job && strpos($job, '-')) {
            // Run from a single job
            $parts = explode('-', $job ?? '');
            $id = $parts[1];
            $service->runJob($id);
            return Command::SUCCESS;
        }
```

if you run  vendor/bin/sake dev/tasks/ProcessJobQueueTask --job=734803, it won't run as it expects the option to have a negative dash, however adding one  (vendor/bin/sake dev/tasks/ProcessJobQueueTask --job=-734803) won't run the code either, as it's in the first position of the job string, and this will evaluate to false.

ideally, we'd want the ability to specify a job ID and have that job run as intended in the code

### How to reproduce

open CLI, run command with vendor/bin/sake dev/tasks/ProcessJobQueueTask --job=XXX

### Possible Solution

        if ($job && is_numeric($job)) {
            // Run from a single job
            $service->runJob($job);
            return Command::SUCCESS;
        }

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-queuedjobs/pull/496

ChloeSartorelli:
PR merged. This will be automatically tagged by GitHub Actions

PR #496, opened 2026-08-10
FIX Job arg in CLI

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-queuedjobs/issues/495

### DIFF
--- a/src/Tasks/ProcessJobQueueTask.php
+++ b/src/Tasks/ProcessJobQueueTask.php
@@ -8,7 +8,6 @@ use SilverStripe\Dev\BuildTask;
 use SilverStripe\PolyExecution\PolyOutput;
 use SilverStripe\PolyExecution\PolyOutputLogHandler;
 use Symbiote\QueuedJobs\Services\AbstractQueuedJob;
-use Symbiote\QueuedJobs\Services\QueuedJob;
 use Symbiote\QueuedJobs\Services\QueuedJobService;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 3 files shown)


src/Tasks/ProcessJobQueueTask.php
```diff
@@ -8,7 +8,6 @@
 use SilverStripe\PolyExecution\PolyOutput;
 use SilverStripe\PolyExecution\PolyOutputLogHandler;
 use Symbiote\QueuedJobs\Services\AbstractQueuedJob;
-use Symbiote\QueuedJobs\Services\QueuedJob;
```
emteknetnz, line 11: Unused import

src/Tasks/ProcessJobQueueTask.php
```diff
@@ -64,11 +63,14 @@ protected function execute(InputInterface $input, PolyOutput $output): int
...
         $job = $input->getOption('job');
-        if ($job && strpos($job, '-')) {
-            // Run from a single job
-            $parts = explode('-', $job ?? '');
-            $id = $parts[1];
-            $service->runJob($id);
+        if ($job) {
+            // Accepts an ID (e.g. "123"), optionally prefixed with anything up to a dash
+            // which covers a job cache file name (e.g. "queuedjob-123") and a leading dash (e.g. "-123")
+            if (!preg_match('#^(?:[^-]*-)?(\d+)$#', $job, $matches)) {
```
emteknetnz, line 69: The old check was `if ($job && strpos($job, '-'))` so it only passed when a dash sat past position 0, and the id then came out of explode('-', $job)[1]. --job=734803 has no dash in it so it fell straight through to a full queue run, and --job=-734803 gives strpos() === 0 which is falsy as well.

tests/ProcessJobQueueTaskTest.php
```diff
@@ -0,0 +1,77 @@
...
+        parent::setUp();
+        Injector::inst()->registerService(new RecordingQueuedJobService(), QueuedJobService::class);
+    }
+
+    public static function provideJobOption(): array
+    {
+        return [
+            'numeric ID' => ['734803', 734803],
+            'job cache file name' => ['queuedjob-734803', 734803],
+            'ID with a leading dash' => ['-734803', 734803],
```
ChloeSartorelli: Please add a scenario with multiple `-`, e.g. `'queued-job-prefix-734803'`

ChloeSartorelli (changes requested):
Looks good, just one test scenario to add for edge-case purposes.

tests/ProcessJobQueueTaskTest.php
emteknetnz: Added that - exposed a issue with regex so I've fixed `ProcessJobQueueTask::execute()` too

Also refactored the dataProvider so it labels the args and splits over multple lines

ChloeSartorelli (approved):
LGTM

